PR #4, opened 2020-01-29
Fail on unknown environment variable

alex-slynko:
The values file is required. 
There is a comment about improving it, but failing fast would be more visible


jspawar:
looks good, thx

Issue #3, opened 2020-01-28
Add namespace annotations during installation 


mike1808:
We at the cf-k8s-networking team need to have additional labels for the metacontroller namespace during installation and was wondering how to do that. So we have several options:

1. Modify the namespace object YAML directly and save it in the repo.
1. Create a namespace object in our configs (not in `_ytt_lib`) and remove one from metacontroller
1. Create an overlay and add labels there.

The reason is to enable Istio automatic sidecar injection and to do it we have to add [`istio-injection=enabled`](https://istio.io/docs/ops/configuration/mesh/injection-concepts/) label to the namespace object. Another solution is to enable injection for **every** namespace and explicitly disable for namespaces when necessary. 

cc @rodolfo2488  @tcdowney  @ndhanushkodi  @rosenhouse

PR #5, opened 2020-01-29
Enable Istio automatic sidecar injection

tcdowney:
We want to enable automatic sidecar injection to have automatic mTLS between cf-for-k8s workloads (both system components and apps). See [our ADR](https://github.com/cloudfoundry/cf-k8s-networking/blob/master/doc/architecture-decisions/0004-strategy-for-securing-network-traffic.md) for a bit more explanation.

As we discussed on Slack, we attempted the `ytt` overlay approach. Happy to pair with y'all on this if you would like it to work differently.

Fixes https://github.com/cloudfoundry/cf-for-k8s/issues/3.

[Story: #170980509](https://www.pivotaltracker.com/story/show/170980509)

Thanks!
@mike1808 && @tcdowney

### DIFF
--- a/config/metacontroller-namespace-overlay.yml
+++ b/config/metacontroller-namespace-overlay.yml
@@ -1,0 +1,9 @@
+#@ load("@ytt:overlay", "overlay")
+
+#@overlay/match by=overlay.subset({"kind":"Namespace", "metadata":{"name":"metacontroller"}})
+---
+metadata:
+  #@overlay/match missing_ok=True
+  labels:
+    #@overlay/match missing_ok=True
+    istio-injection: enabled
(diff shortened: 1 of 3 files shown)


jspawar:
@mike1808 + @tcdowney 

We wanted to suggest some changes to this PR and actually validate it; however, this was merged before the RelInt team could look at it. As a result, we have reverted the merge commit to master. Also, master is now protected from merging from anyone besides the RelInt team, so should hopefully not be an issue going forward.

As for the suggestions we wanted to make: everything in the `config/metacontroller-namespace-overlay.yml` would make better sense to live inside of the `config/networking.yml` file given that metacontroller is only pulled in as a dependency of the networking component. We are trying to keep everything contained to individual component files within the `config` directory of this repo for now.

Thanks,
Jwal + @davewalter

PR #9, opened 2020-02-03
test gitbot integration

jspawar:
testing...


jspawar:
story created; gitbot integration validated. closing out

PR #12, opened 2020-02-05
Add automatic Istio sidecar injection (second attempt!)

tcdowney:
This is a do-over for https://github.com/cloudfoundry/cf-for-k8s/pull/5!

---

We want to enable automatic sidecar injection to have automatic mTLS between cf-for-k8s workloads (both system components and apps). See [our ADR](https://github.com/cloudfoundry/cf-k8s-networking/blob/master/doc/architecture-decisions/0004-strategy-for-securing-network-traffic.md) for a bit more explanation.

As we discussed on Slack, we attempted the `ytt` overlay approach. Happy to pair with y'all on this if you would like it to work differently.

Fixes https://github.com/cloudfoundry/cf-for-k8s/issues/3.

[Story: #170980509](https://www.pivotaltracker.com/story/show/170980509)

Thanks!
@mike1808 && @tcdowney

### DIFF
--- a/config/metacontroller-namespace-overlay.yml
+++ b/config/metacontroller-namespace-overlay.yml
@@ -1,9 +1,0 @@
-#@ load("@ytt:overlay", "overlay")
-
-#@overlay/match by=overlay.subset({"kind":"Namespace", "metadata":{"name":"metacontroller"}})
----
-metadata:
-  #@overlay/match missing_ok=True
-  labels:
-    #@overlay/match missing_ok=True
-    istio-injection: enabled
--- a/config/metacontroller-namespace-overlay.yml
+++ b/config/metacontroller-namespace-overlay.yml
@@ -1,0 +1,9 @@
+#@ load("@ytt:overlay", "overlay")
+
+#@overlay/match by=overlay.subset({"kind":"Namespace", "metadata":{"name":"metacontroller"}})
+---
+metadata:
+  #@overlay/match missing_ok=True
+  labels:
+    #@overlay/match missing_ok=True
+    istio-injection: enabled
--- a/config/networking.yml
+++ b/config/networking.yml
@@ -2,6 +2,7 @@
 #@ load("@ytt:library", "library")
 #@ load("@ytt:template", "template")
 #@ load("@ytt:base64", "base64")
+#@ load("@ytt:overlay", "overlay")
 
 #@ def networking_values():
 systemNamespace: #@ data.values.system_namespace
@@ -26,3 +27,12 @@ cfroutesync:
 --- #@ template.replace(networking.with_data_values(networking_values()).eval())
 --- #@ template.replace(metacontroller.eval())
 --- #@ template.replace(istio.eval())
+
+
+#@overlay/match by=overlay.subset({"kind":"Namespace", "metadata":{"name":"metacontroller"}})
+---
+metadata:
+  #@overlay/match missing_ok=True
+  labels:
+    #@overlay/match missing_ok=True
+    istio-injection: enabled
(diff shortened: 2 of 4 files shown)


cppforlife:
@tcdowney im trying this out locally:

```
$ kapp deploy -a cf -f <(ytt -f config/ -f /tmp/cf-values.yml)
Target cluster 'https://35.239.160.185' (nodes: gke-dk-jan-9-default-pool-a218b1c9-55sl, 4+)

kapp: Error: Expected not to find same resource 'namespace/cf-system (v1) cluster' with different content
```

so checking this out...

```
$ ytt -f config/ -f /tmp/cf-values.yml | grep -5 "name: cf-system"| grep -5 "kind: Namespace"
---
apiVersion: v1
kind: Namespace
metadata:
  name: cf-system
....
---
apiVersion: v1
kind: Namespace
metadata:
  name: cf-system
  labels:
    istio-injection: enabled
```

it seems that capi includes their own copy of cf-system:

```
$ grep -R "name: cf-system" config/_ytt_lib/
config/_ytt_lib//eirini/rendered.yml:          name: cf-system
config/_ytt_lib//eirini/rendered.yml:          name: cf-system
config/_ytt_lib//github.com/cloudfoundry/capi-k8s-release/templates/namespace.yml:  name: cf-system

$ cat config/_ytt_lib//github.com/cloudfoundry/capi-k8s-release/templates/namespace.yml
apiVersion: v1
kind: Namespace
metadata:
  name: cf-system
```

so it seems that we need to either add an overlay to remove capi's cf-system or change our own cf-system namespace. i think ultimately it doesnt make sense for capi to "own" cf-system namespace, hence i would favor first option.

jspawar:
+1 to Dmitriy's suggestion: I think we should we be responsible for the creation of the shared `cf-system` and `cf-workloads` namespaces

tcdowney:
Makes sense, that's something y'all will take care of though right? There's no rush on this PR, just wanted to get it resubmitted.

jenspinney:
Hey @tcdowney @mike1808. When we tried to deploy this (after resolving the trivial merge conflict), we saw errors in the capi-api-server pod in both the `istio-proxy` and `capi-api-server` containers. Both were in a `CrashLoopBackoff` state. Are you all seeing this as well or does it deploy successfully on your setup when merging into the latest on master?

tcdowney:
@jenspinney we believe this error is due to an Istio bug that should be fixed in `1.4.3`.
https://github.com/istio/istio/issues/19096#issuecomment-569080070

We're seeing it here because Cloud Controller has multiple containers in the `capi-api-server` pod and one of them (`cloudfoundry/cloud-controller-ng`) does not declare any `containerPorts`.

```
- image: cloudfoundry/cloud-controller-ng
    imagePullPolicy: Always
    name: capi-api-server
    ...
  - image: nginx:latest
    imagePullPolicy: Always
    name: nginx
    ports:
    - containerPort: 80
      protocol: TCP
    ...
```

The `cloudfoundry/cloud-controller-ng` container doesn't actually listen on any ports so it is configured correctly. We think the correct solution here is to try and upgrade from istio `1.4.2` to `1.4.3`.

acosta11:
Manually merged with an istio bump some additional ordering and config reorganization.
https://github.com/cloudfoundry/cf-for-k8s/commit/08b4cd8a1c3057d8e123f311d4a5d827bd50a536
https://github.com/cloudfoundry/cf-for-k8s/commit/c2b6be42c396f2f5ca22c9b37b6e73e3d0a127d2

Thanks,
Andrew and Dmitriy @cppforlife

PR #13, opened 2020-02-05
remove caBundle empty string configuration to avoid kapp consider it as a change

cppforlife:
caBundle value is populated by the cluster, hence kapp will consider it a change if it is present

i reran build/istio/build.sh to produce updated `config/_ytt_lib/istio/all.yml`


davewalter:
Thanks @cppforlife!

PR #16, opened 2020-02-09
Add instructions to target, verify app, and delete CF

amitkgupta:

Add instructions to target, verify app, and delete CF.

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

1. Follow the original `docs/deploy.md`, see that it fails to push an app because no org/space is targetted. Follow the new `docs/deploy.md` and see that it works.
2. Follow the original docs with one system domain, delete the installation, and follow again with a new system domain; see that it fails because `/tmp/cf-vars.yaml` has certificates for the previous domain. Repeat with the new `docs/deploy.md` and see that it works.

- [x ] **Slightly Less than Urgent**

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -22,7 +22,7 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
 1. Create a "CF Installation Values" file and configure it:
    1. Use `./hack/generate-values.sh <system_domain>` to automatically generate values with [bosh interpolate](https://bosh.io/docs/cli-v2-install/#install)
    ```bash
-   $ ./hack/generate-values.sh cf.example.com > /tmp/cf-values.yml
+   $ ./hack/generate-values.sh <system_domain> > /tmp/cf-values.yml
    ```
    1. Alternatively, create a file called `/tmp/cf-values.yml`. You can use `sample-cf-install-values.yml` in this directory as a starting point.
    1. Change the `system_domain` and `app_domain` to your desired domain address
@@ -38,12 +38,14 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
    of the Istio Ingress Gateway service. You can retrieve the external IP of this service by running
    `kubectl get svc -n istio-system istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[*].ip}'`
    1. If you used the `./hack/generate-values.sh` script then you should only
-      configure a single DNS record for the domain you passed as input to the
-      script and have it resolve to the Ingress Gateway's external IP
+      configure a single DNS record for the wildcard subdomain of the domain
+      you passed as input to the script and have it resolve to the Ingress
+      Gateway's external IP
 
 ## Validate the deployment
 
-1. Set up cf cli to point to CF:
+1. Get the `<cf_admin_password>` with the `bosh interpolate /tmp/cf-values.yml --path /cf_admin_password`
+command and set up cf CLI to point to CF:
    ```bash
    $ cf api --skip-ssl-validation https://api.<system_domain>
    $ cf auth admin <cf_admin_password>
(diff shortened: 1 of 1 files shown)


bin/uninstall-cf.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Delete CF for Kubernetes
+kapp delete -a cf -y
```
pivotaljohn: In fact, we're looking to move _away_ from collecting wrapper scripts.  Would you mind inlining this into the instructions?

docs/deploy.md
```diff
@@ -38,23 +38,62 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
...
+      Gateway's external IP
 
 ## Validate the deployment
 
-1. Set up cf cli to point to CF:
+1. Set up cf CLI to point to CF:
    ```bash
    $ cf api --skip-ssl-validation https://api.<system_domain>
-   $ cf auth admin <cf_admin_password>
+   $ cf auth admin $(bosh interpolate /tmp/cf-values.yml --path /cf_admin_password)
```
pivotaljohn: Let's keep the angle-bracketed line and suggest that the value for `cf_admin_password` can be obtained through the `bosh interpolate` expression.

This keeps with the convention used in the rest of the instructions.

docs/deploy.md
```diff
@@ -38,23 +38,62 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
...
+   CF_INSTANCE_IP: 10.32.2.15
+   CF_INSTANCE_PORT: "8080"
+   CF_INSTANCE_PORTS: '[{"external":8080,"internal":8080}]'
+   HOME: /home/some_docker_user
+   HOSTNAME: diego-docker-app-test-space-2a80ac86bf-0
+   KUBERNETES_PORT: tcp://10.0.16.1:443
+   ...
+   VCAP_APP_HOST: 0.0.0.0
+   VCAP_APP_PORT: "8080"
+   VCAP_SERVICES: '{}'
```
pivotaljohn: While this provides much nicer output, it is probably more involved than the typical consumer of these instructions at this junction.  We're looking to "just" verify that the deploy was successful... even ugly JSON response is okay for that purpose.

docs/deploy.md
```diff
@@ -38,23 +38,62 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
...
+   ```
+
+## Delete the deployment
+
+1. Run the uninstall script
+   ```
+   $ ./bin/uninstall-cf.sh
```
pivotaljohn: See request to inline, from above.

docs/deploy.md
```diff
@@ -38,23 +38,62 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
...
+   ```
+
+1. Clean up generated variables so that the next time you install from scratch, it runs cleanly:
+   ```
+   $ ./hack/cleanup-vars.sh
+   ```
```
pivotaljohn: Depending on the user's need, this may or may not be a desired outcome.  Let's drop this (and the script) for now.

docs/deploy.md
amitkgupta: thanks @pivotaljohn, what do you recommend? Are you recommending to delete this section?

paulczar: I think just telling the user to expect a JSON document containing env variables would be enough.

docs/deploy.md
```diff
@@ -38,12 +38,14 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
...
+      Gateway's external IP
 
 ## Validate the deployment
 
-1. Set up cf cli to point to CF:
+1. Get the `<cf_admin_password>` with the `bosh interpolate /tmp/cf-values.yml --path /cf_admin_password`
+command and set up cf CLI to point to CF:
    ```bash
    $ cf api --skip-ssl-validation https://api.<system_domain>
    $ cf auth admin <cf_admin_password>
```
paulczar: could you set this line to be `cf auth admin $(bosh interpolate /tmp/cf-values.yml --path /cf_admin_password)` and reduce the need to copy/paste or type ?

docs/deploy.md
```diff
@@ -22,7 +22,7 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
 1. Create a "CF Installation Values" file and configure it:
    1. Use `./hack/generate-values.sh <system_domain>` to automatically generate values with [bosh interpolate](https://bosh.io/docs/cli-v2-install/#install)
    ```bash
-   $ ./hack/generate-values.sh cf.example.com > /tmp/cf-values.yml
+   $ ./hack/generate-values.sh <system_domain> > /tmp/cf-values.yml
```
paulczar: maybe if we tell the user to set `SYSTEM_DOMAIN=<system_domain>` at the start, then use that variable in all of the subsequent commands, reducing the copy/paste work.

docs/deploy.md
amitkgupta: I agree @paulczar, that's what I had in my original PR and was asked to switch it to this! Happy to PR whatever the maintainers collectively agree upon, my vote is for the same you're suggesting. Please let me know what y'all agree to.

docs/deploy.md
amitkgupta: @paulczar ^^

docs/deploy.md
pivotaljohn: @amitkgupta, just noting that a valid response is a blob of JSON seems enough for this task. 👍

docs/deploy.md
pivotaljohn: We're not sure the lifecycle of this script.  Let's not invest more than needed.  Also, we're using the angle brackets typography in this document.  Agreed that this means a little more typing from the end-user, right now.

vitreuz:
Hey @amitkgupta,

It's been a while since this PR was active, we're going to go ahead and close it. Please feel free to reopen it when you have time to address the issues brought up by @pivotaljohn. Thanks

PR #20, opened 2020-02-11
Delete cf org only if it is created

alex-slynko:

Delete organisation only if it is created

I used a wrong password and the test was running for a long time.

- [ ] YES
- [X] NO

- [X] **Slightly Less than Urgent**


pivotaljohn:
LGTM.  Thank you!

PR #21, opened 2020-02-11
docs: clarify that the BOSH cli is required for generating values

tcdowney:

My pair and I were deploying cf-for-k8s this morn and were shocked (🤯) to discover we needed to have the `bosh` (BOSH Outer Shell) CLI  installed for generating values. Can we note this in the requirements? Thanks!

https://github.com/cloudfoundry/cf-for-k8s/blob/73f18c5733b25e291a0a4bd7d4dd4bfeeda43032/hack/generate-values.sh#L66-L70

- [x] YES
- [] NO

- [ ] YES - please specify
- [x] NO

### Pair
@ndhanushkodi 

## Alternative Solutions
You could also add `ytt interpolate`. 👍

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -27,7 +27,7 @@ Make sure that your Kubernetes config (e.g, `~/.kube/config`) is pointing to the
    You have the option of auto-generating the installation values or creating the values by yourself. 
 
    #### Option 1 - Generate the install values
-   The script relies on bosh [interpolate](https://bosh.io/docs/cli-v2-install/#install) to generate the install values
+   **NOTE:** The script relies on [bosh interpolate](https://bosh.io/docs/cli-v2-install/#install) to generate the install values
    ```bash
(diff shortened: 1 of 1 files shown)


jenspinney:
@tcdowney we just merged a PR from @Syerram yesterday (#18) that I think may have helped this a little. His PR makes it more clear that are two ways to generate those values. Only one of those ways requires BOSH, so I wonder if we could instead put the BOSH CLI requirement closer to where we say you can use `hack/generate-values.sh`?

tcdowney:
@jenspinney thanks for taking a look! That commit by @Syerram helps a lot. We tweaked the PR a bit just to make that existing line more obvious.

Thanks!

Or you could close this, it's not super necessary anymore. 😅

cwlbraa:
I read this script shortly after it got added, and I'd just like to say that I enjoyed it immensely.

PR #26, opened 2020-02-13
fix: put CF vars to unique temp dir

mike1808:
* `bosh interpolate` reuses vars from varstore file if the file exists
so if you reuse this scipt for a new environment the certificate
generation won't happen and they will contain old domains

Put CF vars to unique temp dir

`bosh interpolate` reuses vars from varstore file if the file exists
    so if you reuse this scipt for a new environment the certificate
    generation won't happen and they will contain old domains

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

Redeploy using hack scripts and check that system components can communicate with UAA. Or check the certificates that SAN field contains the right domain.

cc @ndhanushkodi

### DIFF
--- a/hack/generate-values.sh
+++ b/hack/generate-values.sh
@@ -8,8 +8,8 @@ set -eu
 # Make sure bosh binary exists
 bosh --version >/dev/null
 
-VARS_FILE=/tmp/cf-vars.yaml
 DOMAIN=$1
+VARS_FILE="/tmp/${DOMAIN}/cf-vars.yaml"
 
 bosh interpolate --vars-store=${VARS_FILE} <(cat <<EOF
 variables:


hack/generate-values.sh
```diff
@@ -8,7 +8,7 @@ set -eu
 # Make sure bosh binary exists
 bosh --version >/dev/null
 
-VARS_FILE=/tmp/cf-vars.yaml
+VARS_FILE="$(mktemp -d)/cf-vars.yaml"
```
jenspinney: I think I remember some inconsistencies across OS's when using `mktemp -d`. I've tested the following out on Mac and Linux and it has the benefit of making it more obvious where this temp dir came from:

```suggestion
VARS_FILE="$(mktemp -d -t 'cf-for-k8s.XXX')/cf-vars.yaml"
```

One side effect of using a different directory every time is that if you wanted to reuse the same certs and passwords from a previous deploy (i.e., you're upgrading instead of deleting and redeploying) you would no longer be able to simply rerun the script to generate any new needed variables. Personally, I think that's ok.

mike1808: `-t` is different in macOS and Linux, for sure. 

> One side effect of using a different directory every time is that if you wanted to reuse the same certs and passwords from a previous deploy

Maybe, we can use `$DOMAIN` and put the var store in `/tmp/${DOMAIN}/cf-vars.yml`?

jenspinney: Oh yeah, I like the approach of using DOMAIN like that.  Just some bugs in the current approach that I'll make comments on directly.

hack/generate-values.sh
```diff
@@ -2,13 +2,15 @@
 
 set -eu
 
+: "${DOMAIN:?}"
```
jenspinney: We first set DOMAIN down below on line 14

hack/generate-values.sh
```diff
@@ -2,13 +2,15 @@
...
+
 ## Usage: ./generate-values.sh <my-domain>
 ## where my-domain will be used as both the system domain and app domain.
 
 # Make sure bosh binary exists
 bosh --version >/dev/null
 
-VARS_FILE=/tmp/cf-vars.yaml
+VARS_FILE="/tmp/${DOMAIN}/cf-vars.yaml"
```
jenspinney: We'll have to `mkdir -p /tmp/${DOMAIN}` sometime before this line.

zankich:
How would people feel about following a more unix philosophy style and writing the contents of values file to STDOUT? Leave it up to the user where they want to place this file.

jenspinney:
@zankich the output of the script does go to STDOUT. The problem is an intermediate file that's used by the BOSH CLI to store the generated credentials. Not sure if there's a way around that in the BOSH CLI.

zankich:
Got it, thanks for the clarification.

hack/generate-values.sh
mike1808: @jenspinney moved it to the next line, so the domain is available.

jenspinney:
Looks good, thanks for the improvement!

PR #30, opened 2020-02-19
add cf-auth-proxy to log-cache for security

chentom88:
Signed-off-by: Melena Suliteanu <msuliteanu@pivotal.io>

We added uaa authorization to log cache access. When accessing log cache, a client will now need to provide a UAA token which has access to the source id being requested.

https://www.pivotaltracker.com/story/show/170084847

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

`cf tail <source_id>` should fetch the logs for the provided source_id if the logged in user has access to source_id. Otherwise, the command should fail.

@pianohacker


config/logging.yml
```diff
@@ -27,14 +27,32 @@ log_cache_gateway:
...
 
 #@ logging = library.get("github.com/cloudfoundry/cf-k8s-logging")
 --- #@ template.replace(logging.with_data_values(logging_values()).eval())
 
+---
+apiVersion: v1
+kind: Secret
+type: Opaque
+metadata:
+  name: system-ca
```
cppforlife: `system-ca` is way too of a generic name for a secret. i would recommend to either make secret name configuration via one of the data values, or hard coding to something more apparent like log-cache-system-ca.

config/logging.yml
```diff
@@ -27,14 +27,32 @@ log_cache_gateway:
...
+  namespace: #@ data.values.system_namespace
+data:
+  tls.crt: #@ data.values.system_certificate.crt
+
 #@overlay/match by=overlay.subset({"kind":"Service", "metadata":{"name": "log-cache"}})
```
cppforlife: since you are bumping log-cache why dont we also fix http -> https change so that this overlay doesnt have to be here?

config/logging.yml
Benjamintf1: This *isn't* the log-cache-system ca, it's the ca used to sign both uaa and capi. I hear there might be a future where uaa and capi have their own ca/certs, I'm open to changing it then.

config/logging.yml
Benjamintf1: Sure, we can look into doing that.

Benjamintf1:
We see the problem here, we're looking at fixing our dns to use services rather then require dns to be set up for the load balancer.

config/logging.yml
Benjamintf1: We could make the secret name configurable though, that's possibly a good move.

cppforlife:
> We see the problem here, we're looking at fixing our dns to use services

yes please!

config/logging.yml
cppforlife: > This isn't the log-cache-system ca

it's a ca that log-cache consumes though, hence i think making it more "injectable" is better (ie yall had to add it as a Secret resource.)

>  I hear there might be a future where uaa and capi have their own ca/certs

longer term (which hopefully happens soon) we ll be relying on istio to do all tls wrapping/unwrapping so there wouldnt be CAs to configure between components (ie no system ca)

Benjamintf1:
There, we fixed it!

pivotaljohn:
@cloudfoundry/cf-loggregator — we've granted your team `write` access to this repository.

From now on, our Smoke Test checks will no longer run on PRs based on forks; they need to be from branches in this repo to get this extra validation.

@chentom88 + @MasslessParticle: 
For _this_ PR, could you please move this work onto such a branch?  AFAIK, you can't _move_ a PR from a forked source, so we'd close this PR and reference it from a new one you'd create from that branch.

p.s. this is the subject which we pinged you in [#cf-for-k8s](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1582573083015800).

pivotaljohn:
Spoke with @MasslessParticle.  He said they'll resubmit on a new PR.  Thank you!

PR #33, opened 2020-02-21
remove reliance on BOSH cli

paulczar:

BOSH cli is only used to do some basic work like creating random strings for
passwords, and generating certs.  This is something we can already do with
linux tools.

By splitting the password and cert generation into separate scripts/ytt values
we show the user some of the power of ytt in loading values from multiple places.

scripts and deploy.md have been updated to pass lint tests.

ci dockerfiles and pipelines have been modified to not install or use bosh cli

a values path has been added to give a user a place to safely put resulting values.yaml
files with a suitable .gitignore to ensure they don't get saved back up into git. This
helps protect the use from the ephemeral nature of tmpfs.

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO


docs/deploy.md
```diff
@@ -2,14 +2,18 @@
...
 You need the following CLIs on your system to be able to run the script:
 
 * [`kapp`](https://k14s.io/#install)
 * [`ytt`](https://k14s.io/#install)
-* Kubernetes cluster requirements
-  * Version 1.14 or higher
-  * A minimum of 5 nodes
-  * A minimum of 2 CPU, 7.5BGB memory per node
+
+> Note: the `bosh` CLI is only used to reduce toil by heing to generate some random passwords and valid self-signed certificates. You can skip it if you're allergic to `bosh`.
```
tybritten: "helping"

paulczar:
This fails CI because it modifies the install script (and the corresponding pipeline jobs) but it appears the pipelines aren't updated in concourse based on the PR, so the old pipeline is calling the updated script ... and failing.

PR #35, opened 2020-02-21
feat: ensure traffic on the mesh is encrypted

XanderStrike:

all unencrypted requests on the mesh will now be rejected

there are two exceptions:
  - capi requires an exception so that its db migrations in an init
  container can still access the db before the sidecar is up
  see: https://github.com/cloudfoundry/capi-k8s-release/issues/12
  - log cache requires an exception because it terminates its own tls in
  a way incompatible with the sidecars

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

1. Deploy cf for k8s
2. Enable docker support 
3. Push an app `cf push httpbin -o cfrouting/httpbin8080` 
4. Identify your service `kubectl get service -n cf-workloads -l cloudfoundry.org/app_guid=$(cf app httpbin --guid)` 
  * Looks like s-GUID 
5. SSH into one of your gateways 
  * `kubectl exec -it $(kubectl get pods -n istio-system | grep ingressgateway | head -n1 | awk '{print $1}') -n istio-system /bin/bash` 
6. Go to `/etc/certs` 
7. Send a plain text request to your app 
  * `curl http://<SERVICE NAME>.cf-workloads.svc.cluster.local:8080` 
  * It should fail with something like "Connectino reset by peer" 
  * NOTE: this doesn't work right now because we have to keep things in PERMISSIVE for CAPI to be HAPI 
8. Send an encrypted and authenticated request by specifying the cert, key, and rootca 
  * `curl -k --cacert root-cert.pem --key key.pem --cert cert-chain.pem https://<SERVICE NAME>.cf-workloads.svc.cluster.local:8080` 
  * See that it succeeds!"

@tcdowney @rodolfo2488 @rosenhouse @ndhanushkodi @keshav-pivotal

### DIFF
--- a/config/postgres.yml
+++ b/config/postgres.yml
@@ -49,3 +49,17 @@ data:
     psql -U postgres -d (@= uaadb.name @) -c "CREATE EXTENSION citext"
 
 --- #@ template.replace(overlay.apply(library.get("postgres").eval(), add_cf_db_namespace()))
+---
+apiVersion: "authentication.istio.io/v1alpha1"
+kind: "Policy"
+metadata:
+  name: "cf-db-allow-plaintext"
+  namespace: "cf-db"
+spec:
+  targets:
+  - name: cf-db-postgresql
+  - name: cf-db-postgresql-headless
+  peers:
+  - mtls:
+      mode: PERMISSIVE
+
--- a/config/postgres.yml
+++ b/config/postgres.yml
@@ -49,7 +49,8 @@ data:
     psql -U postgres -d (@= uaadb.name @) -c "CREATE EXTENSION citext"
 
 --- #@ template.replace(overlay.apply(library.get("postgres").eval(), add_cf_db_namespace()))
----
+--- #! explanation: the capi-db's sidecar needs to accept plain text connections from capi's init container.
+    #! see https://github.com/cloudfoundry/capi-k8s-release/issues/12
 apiVersion: "authentication.istio.io/v1alpha1"
 kind: "Policy"
 metadata:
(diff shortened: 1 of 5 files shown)


config/postgres.yml
```diff
@@ -49,3 +49,17 @@ data:
...
+metadata:
+  name: "cf-db-allow-plaintext"
+  namespace: "cf-db"
+spec:
+  targets:
+  - name: cf-db-postgresql
+  - name: cf-db-postgresql-headless
+  peers:
+  - mtls:
+      mode: PERMISSIVE
```
cppforlife: is this set to permissive because of the capi's init container?

XanderStrike: Yes exactly.

cppforlife: would you mind adding a comment in this file mentioning that?

config/postgres.yml
```diff
@@ -49,3 +49,17 @@ data:
...
+metadata:
+  name: "cf-db-allow-plaintext"
+  namespace: "cf-db"
+spec:
+  targets:
+  - name: cf-db-postgresql
+  - name: cf-db-postgresql-headless
+  peers:
+  - mtls:
+      mode: PERMISSIVE # this allow's CAPI's init container to reach the database before the sidecar is up
```
cppforlife: has to be '#!' instead of '# ' as ytt will complain about ambiguous comment

config/postgres.yml
XanderStrike: We've added comments to both the capi and log-cache workaround configuration.

config/postgres.yml
XanderStrike: We changed it after seeing your edit.

jenspinney:
Looks good. Thanks!

XanderStrike:
Thank you!

pivotaljohn:
BTW, fantastic acceptance steps!!!  Thank you for taking the time to do so... and so clearly. 👍

PR #39, opened 2020-02-24
introduce cf-auth-proxy without tls

MasslessParticle:
[#171288061]

Signed-off-by: Travis Patterson <tpatterson@pivotal.io>

Adds the CF Auth Proxy so requests to Log Cache are authenticated. Also, remove TLS from CF Log Cache so istio can handle it.

https://www.pivotaltracker.com/story/show/171288061

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

- Install CF-for-k8s
- Authenticate to cf
- curl `curl -H "Authorization: $(cf oauth-token)" -k http://log-cache.<system_domain>/api/v1/meta`

@chentom88 @pianohacker


pivotaljohn:
@Benjamintf1.  We noticed that there are a couple of `cf-rel-int-status-bot` comments that were deleted (GitHub says it was from your account).

In general, we'd prefer that these things be simply explained.  Even a "ignore those messages, we got this." is better than deleting comments.

Also, we rely on the signal that git-Tracker integration will drop a label on the corresponding story in our backlog that there is "new activity".  This only happens when comments are added, not deleted... which means an issue could (accidentally) get ignored because we believe we're waiting on the contributor for some reason.

config/logging.yml
```diff
@@ -27,28 +27,21 @@ log_cache_gateway:
...
   namespace: #@ data.values.system_namespace
-spec:
-  targets:
-  - name: log-cache
-  peers:
-  - mtls:
-      mode: PERMISSIVE
-
+data:
+  tls.crt: #@ data.values.system_certificate.crt
```
pivotaljohn: Curious...  why are we referencing/trusting the certificate, directly... rather than trusting the Certificate Authority (that was used to sign the cert)?

config/_ytt_lib/github.com/cloudfoundry/cf-k8s-logging/config/manifests/500-log-cache-deployment.yaml
```diff
@@ -18,6 +18,37 @@ spec:
...
+          - containerPort: 8083
+        env:
+        - name: LOG_CACHE_GATEWAY_ADDR
+          value: https://localhost:8081
+        - name: CAPI_CA_PATH
+          value: /system-ca/tls.crt
+        - name: UAA_CA_PATH
+          value: /system-ca/tls.crt
+        - name: CAPI_ADDR
+          value: #@ "https://capi.{}.svc.cluster.local:80".format(data.values.system_namespace)
```
pivotaljohn: A little confused here:
- the protocol of `CAPI_ADDR` is http*s*
- the port of `CAPI_ADDR` is 80 — usually reserved for plaintext HTTP
- looking at the API server, the `capi` Service names port 80 as http.  We assume the API server is actually serving plaintext on `:80` (https://github.com/cloudfoundry/capi-k8s-release/blob/98b34698e61332be8fa7955d793cd60880208ed2/templates/service.yml#L10-L13)
- We're talkin' plaintext to UAA, but TLS to CAPI?!

Could you help me understand?

Benjamintf1: We tried using https for both, but uaa didn't accept https. We saw capi_addr was 80 on the service, so we used that. https://github.com/cloudfoundry/cf-for-k8s/blob/409efb3bbe6df6bd6e53ee71aa2ad9fac30e5b0c/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/service.yml#L10. I'm not sure why capi actually accepts(or seems to accept https), but it does.

config/logging.yml
Benjamintf1: We should probably be trusting the ca, this is totally a mistake. Although, I suspect if we switch capi over to http(if that's desireable), we could get rid of this certificate in general.

pivotaljohn: Cool.

And yes, to clarify: it's absolutely desirable that _all_ components serve plaintext and allow the Istio sidecars to handle TLS/mTLS.

config/_ytt_lib/github.com/cloudfoundry/cf-k8s-logging/config/manifests/500-log-cache-deployment.yaml
pivotaljohn: Got it... thank you for that explanation, @Benjamintf1.

I'm parsing your comment precisely and reading that CAPI might well serve HTTP, but y'all hadn't tried (expecting, at this time, to be trying to make this communication encrypted)... is that right?

Either way, we've been assured by CF for K8s Networking that Istio has been configured to encrypt traffic between services.  You should be able to serve and consume plaintext.

config/logging.yml
pivotaljohn: ... and to be clear, I'm talking about so-called "east-west" traffic (i.e internal).

config/_ytt_lib/github.com/cloudfoundry/cf-k8s-logging/config/manifests/500-log-cache-deployment.yaml
Benjamintf1: yup, we hadn't attempted that yet.

Benjamintf1:
We updated the pr as per suggestions, no more certs/etc.

config/_ytt_lib/github.com/cloudfoundry/cf-k8s-logging/config/manifests/500-log-cache-deployment.yaml
Benjamintf1: We did it

config/logging.yml
Benjamintf1: We removed the cert entirely.

Benjamintf1:
We saw there were conflicts, these conflicts seemed to cause problems when we used the github ui to fix conflicts(github lies). So we rebased instead.

PR #41, opened 2020-02-25
feat: use SDS for ingressgateway cert distribution

tcdowney:
- this will allow for easier rotation of certificates
- https://istio.io/docs/tasks/traffic-management/ingress/secure-ingress-sds/

We want to use Envoy's [Secret Discovery Service (SDS)](https://istio.io/docs/tasks/traffic-management/ingress/secure-ingress-sds/) functionality to make it easier for operators to dynamically add/rotate certs on the platform.

cf-k8s-networking story: [#171238418](https://www.pivotaltracker.com/story/show/171238418)

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

After this change you can observe that the `cf-system` `Gateway` is loading its certificates by the credential using SDS instead of referencing volume mounted paths.

```
k -n cf-system get gateways.networking.istio.io istio-ingressgateway -o yaml | grep credentialName | tail -n 1
```

You can also see that an additional container called `ingress-sds` is running on the `istio-ingressgateway` pod.

```
k -n istio-system get pod istio-ingressgateway-675757687f-p52zf -o jsonpath='{.spec.containers[*].name}'
ingress-sds istio-proxy
``` 

After you've confirmed that the configuration is in place. Just interact with the cluster normally using the `cf cli` using `https`.

e.g.
```
cf curl /v3/apps -v
```

@rodolfo2488 @XanderStrike @rosenhouse @ndhanushkodi @keshav-pivotal @Gerg

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml
+++ b/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml
@@ -5923,6 +5923,29 @@ spec:
                 - ppc64le
                 - s390x
       containers:
+      - env:
+        - name: ENABLE_WORKLOAD_SDS
+          value: "false"
+        - name: ENABLE_INGRESS_GATEWAY_SDS
+          value: "true"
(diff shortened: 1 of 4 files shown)


rodolfo2488:
lgtm :jots-down:

PR #53, opened 2020-03-04
Bump UAA to v74.15.0

acosta11:

Bump UAA to the latest release v74.15.0 and remove the now-redundant service overlay.

Thread in slack regarding how often we bump UAA.
https://pivotal.slack.com/archives/C3LUD2L04/p1582826401017800

- [X] YES
- [ ] NO

- [ ] YES - please specify
- [ X] NO

Deploy and smoke tests continue to pass.

- [X] **Urgent** - unblocks current or future work

cc @davewalter and @jamespollard8 

### Notes
We ended up having to manually update the vendir lockfile because github was rate limiting us trying to run the normal vendir sync command.


config/_ytt_lib/github.com/cloudfoundry/uaa/k8s/templates/values/version.yml
```diff
@@ -0,0 +1,3 @@
+#@data/values
+---
+version: "74.12.0"
```
acosta11: Side note: @shamus and @dbeneke we noticed a potential version mismatch while pulling in v74.15.0 . It looks like the value was added fairly recently but isn't automatically updated.

joshuatcasey: Thanks! We've got a [story coming up](https://www.pivotaltracker.com/story/show/169720407) that should force us to fix this issue...we've also brought it up with our PMs to surface that it is currently causing pain.

PR #68, opened 2020-03-17
Support "no loadbalancer" configuration on

XanderStrike:
🚫 This must be merged after #66  🚫

This change deploys istio ingress gateways as a daemonset to all nodes, and configures them to listen on port 80 and 443. This way any user can configure their hosts/dnsmasq/etc to point at any given node and have the full functionality of the cluster.

Also, in production environments, operators can configure hardware load balancers to balance between all of their node ips.

https://www.pivotaltracker.com/story/show/171067809

### Have you read the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/docs/contributing.md)?

- [X] YES
- [ ] NO

- [ ] YES - please specify
- [X] NO

WHEN I have created a cf-for-k8s environment with domain `*.example.com`
AND I have modified my `/etc/hosts` to point `api.example.com` to one of my nodes
THEN I am able to `cf login -a api.example.com` and all other fun cf commands

WHEN I have created a cf-for-k8s environment with domain `*.example.com`
AND I have pushed an app `bananas`
AND I have modified my `/etc/hosts` to point `bananas.example.com` to one of my nodes
THEN I am able to curl `bananas.example.com` and see great success

- [X] **Urgent** - unblocks current or future work

@ndhanushkodi @XanderStrike @kauana @christianang @tcdowney @rodolfo2488 @rosenhouse

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml
+++ b/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml
@@ -5209,7 +5209,7 @@ spec:
         env:
         - name: CITADEL_ENABLE_NAMESPACES_BY_DEFAULT
           value: "true"
-        image: docker.io/istio/citadel:1.4.3
+        image: docker.io/istio/citadel:1.4.5
         imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:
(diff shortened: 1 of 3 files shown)


rosenhouse:
> This change deploys istio ingress gateways as a daemonset to all nodes

Looking at the YAML diff, it appears that this change doesn't do anything to the istio ingress gateways, but instead adds a new set of Envoys, which statically forward to the istio ingress gateways.

XanderStrike:
Ah my mistake, the story we picked up didn't have that context and we assumed doing what was in the doc was sufficient. I'll reject the story and close while we work out how to do that instead.

PR #72, opened 2020-03-19
Bump capi-k8s-release with various enhancements

selzoc:

🚫 We only accept PRs to `develop` branch. If this is an exception, please specify why 🚫

- Fix `log_cache` property so logs work via the `cf` cli [see
https://github.com/cloudfoundry/capi-k8s-release/pull/18]
- Remove the ability to configure the kubernetes api url - since we're
running in the cluster itself, we are fine using
`https://kubernetes.default`
  - No longer need to look at current kubeconfig to determine kubernetes
  api url, so removed that section of `hack/generate-values.sh`
- Enable `metric-proxy` communication instead of logcache for metrics
usage [see https://github.com/cloudfoundry/capi-k8s-release/pull/16]

https://www.pivotaltracker.com/story/show/171735462
https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1584553698495500

### Have you read the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/docs/contributing.md)?

- [x] YES
- [ ] NO

- [ ] YES - please specify
- [x] NO

- You should be able to see logs for an app
- You should no longer need/be able to configure the kubernetes api url

@cloudfoundry/cf-capi

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/api_server_deployment.yml
+++ b/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/api_server_deployment.yml
@@ -44,6 +44,12 @@ spec:
           #@ if/end data.values.eirini.serverCerts.secretName:
           - name: eirini-certs
             mountPath: /config/eirini/certs
+          #@ if/end data.values.metric_proxy.cert.secret_name:
+          - name: metric-proxy-certs
+            mountPath: /config/metric_proxy/certs
+          #@ if/end data.values.metric_proxy.ca.secret_name:
+          - name: metric-proxy-ca
+            mountPath: /config/metric_proxy/ca
         - name: capi-local-worker
           image: #@ data.values.images.ccng
           imagePullPolicy: Always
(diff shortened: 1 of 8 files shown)


acosta11:
Manually verified cf logs was working as intended with cf cli v6.50 because our ci currently only uses v6.49 which lacks native log cache integration for the cf logs command. Thanks for the PR, looks good!

loewenstein:
We would be interested to separate the clusters where capi runs and where applications run, maybe not now but at a later point.
What exactly is capi doing with the now hard-coded local api server? Is that for creating kpack CRs? Is there any chance apps run custom code as part of being staged by kpack? Otherwise, running kpack where capi runs is probably fine. 

Are there plans to do more than kpack with the local api server?

selzoc:
Currently, the cloud controller only talks to k8s to create kpack crs.  In the future, we'll probably also be querying k8s to list out available buildpacks, and also creating other kpack crs to support buildpack admin flows.

In addition, we're starting work soon with the networking team to create route crs.

I think we're also interested in exploring multi-cluster deployments in the future, in which case having just one k8s api url will be insufficient anyway.  So for now I just removed the configurability to reduce the amount of things folks have to configure to get up and running with cf-for-k8s.

loewenstein:
> In addition, we're starting work soon with the networking team to create route crs.

Is there already something available on what exactly you plan to do with regards to route CRs? 

Anything else planned to be modeled in the k8s api server instead of the ccdb?

selzoc:
@loewenstein I wouldn't say that it's an "instead" at this point.  The ccdb remains the source of truth for the desired state of routes.  This seems like we're going pretty far afield of this pr discussion now, but feel free to reach out on slack.

I'll leave you with this doc on the proposed route CRD: https://docs.google.com/document/d/1DF7eTBut1I74w_sVaQ4eeF74iQes1nG3iUv7iJ7E35U/edit#heading=h.1z6hvc8yon6k

Issue #22, opened 2020-02-12
Kubernetes >= 1.16.x


jseely:
Error Message:
```
kapp: Error: Expected to find kind 'extensions/v1beta1/DaemonSet', but did not 
```
Related to deprecation of `extensions/v1beta1/DaemonSet`: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

kramerul:
We have the same problem. The title should be "kubernetes >= 1.16.x". After changing `config/_ytt_lib/github.com/cloudfoundry/cf-k8s-logging/config/manifests/500-fluentd-daemonset.yaml` accordingly we ran into #23

julian-hj:
FWIW, this markup for the fluentd DaemonSet appears to work on newer k8s versions...it just needs a selector added in addition to changing the `apiVersion`

```
---
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: fluentd
  namespace: cf-system
  labels:
    app: fluentd
    version: v1
  annotations:
    kapp.k14s.io/change-rule.istio-sidecar-injector: upsert after upserting cf-for-k8s.cloudfoundry.org/istio-sidecar-injector
spec:
  selector:
    matchLabels:
      app: fluentd
      version: v1
  template:
    metadata:
      labels:
        app: fluentd
        version: v1
    spec:
      serviceAccountName: fluentd-service-account
      tolerations:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
      imagePullSecrets:
      - name: image-registry-credentials
      containers:
      - name: fluentd
        image: logcache/cf-k8s-logging
        resources:
          limits:
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 200Mi
        volumeMounts:
        - name: fluentd-config
          mountPath: /fluentd/etc
        - name: varlog
          mountPath: /var/log
        - name: varvcapstore
          mountPath: /var/vcap/store
        - name: dockercontainers
          mountPath: /var/lib/docker/containers
        - mountPath: /ca
          name: ca
          readOnly: true
      terminationGracePeriodSeconds: 30
      volumes:
      - name: fluentd-config
        configMap:
          name: fluentd-config
      - name: varlog
        hostPath:
          path: /var/log
      - name: varvcapstore
        hostPath:
          path: /var/vcap/store
      - name: dockercontainers
        hostPath:
          path: /var/lib/docker/containers
      - name: ca
        secret:
[...]

PR #87, opened 2020-03-25
bump cf-k8s-logging to 0.1.0

Benjamintf1:
- update images to ubuntu
- update to be able to read from containerd
- updating logs to be less latent
- pin images based on release

---

@pianohacker @MasslessParticle


jamespollard8:
this fixes #22

jamespollard8:
Looks great - thanks @Benjamintf1!. We confirmed that this works well on GKE with Kubernetes version 1.16.

In general we do find it quite helpful if you could provide acceptance criteria so that we know how to validate that the change is working as expected. But we requested this update and have sufficient context this time. 👍

PR #89, opened 2020-03-26
Add dockerhub kpack configuration to generate-values.sh

zankich:
> Add dockerhub support to generate-values.sh

---

- Let us know if getting this merged is urgent.

_Special notes for your reviewer /  **Acceptance Criteria** (if applicable)_

> **_Example Acceptance Criteria:_**
> For a PR that contains a new Deployment resource definition
**When** I run `kubectl get deployment <NAME_OF_DEPLOYMENT_RESOURCE>`
**Then** I see that the new deployment exists

_Resources_

> _Include any links to other PRs, stories, slack discussions, etc... that will help establish context._

### DIFF
--- a/hack/generate-values.sh
+++ b/hack/generate-values.sh
@@ -17,6 +17,12 @@ flags:
       (optional) Filepath to the GCP Service Account JSON describing a service account
       that has permissions to write to the project's container repository.
 
+  -du, --dockerhub-username
+      (optional) dockerhub basic auth username.
+
+  -dp, --dockerhub-password
+      (optional) dockerhub basic auth password.
+
 EOF
   exit 1
 }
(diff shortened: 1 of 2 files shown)


cppforlife:
i think it's better to just document that user can go ahead and add stuff to values.yml (ultimately that is the API to the user for cf-for-k8s). arguably we should delete GKE integration as well. would love to chat more with cf-for-k8s team.

ericpromislow:
> i think it's better to just document that user can go ahead and add stuff to values.yml (ultimately that is the API to the user for cf-for-k8s). arguably we should delete GKE integration as well. would love to chat more with cf-for-k8s team.

There are two "personas" we are serving with this product.  The primary, of course, are the Platform Engineers (aka "Alana"s).  The other are those contributing to the product (i.e. "Rachel"s).

Anything in the `hack` directory are artifacts primarily serving as scaffolding for moving the product forward.  Today, that means making it as easy as possible for Rachels to integrate into the latest revision of `cf-for-k8s`.

At a later point, we can (and expect to, in many cases) remove support for these options and give users step-by-step instructions for how to achieve that configuration (either in `values.yml` and/or ytt templating).  In the meantime, this is a good indication of what our customers need.

Please feel free to swing by for a conversation about this.

cppforlife:
> Please feel free to swing by for a conversation about this.

would be happy to. i happened to chat with dave and andrew today about few things including this, so might be good to have a larger convo.

hack/generate-values.sh
```diff
@@ -256,3 +285,15 @@ kpack:
...
 EOF
 fi
+
+if [[ -n "${DOCKERHUB_USERNAME:=}" ]]; then
+  cat <<EOF
+
+kpack:
+  registry:
+    hostname: https://index.docker.io/v1/
+    repository: ${DOCKERHUB_USERNAME}
```
davewalter: Is there ever a case where the operator would want to specify a repository that is not the same as the username?

cppforlife:
i dont actually see where `data.values.app_registry.repository` is used.

PR #92, opened 2020-03-26
Integrates metric-proxy to return app metrics

kkburr:
Signed-off-by: Caitlyn Yu <cyu@pivotal.io>
Signed-off-by: David Timm <dtimm@pivotal.io>

This uses the [metrics-proxy](https://github.com/cloudfoundry/metric-proxy) to return metrics for `cf push` and `cf app`. The metric-proxy makes a request to the Kube API and formats the results to match the existing log-cache API.

**_Acceptance Criteria:_**

**When** I run `kubectl get pod -n cf-system | grep metric-proxy`
**Then** I see the metric-proxy pod is running

**When** I run `cf push <my-app>` ie `cf app <my-app>`
**Then** I see cpu and memory metrics listed in the output.

For more information, see our [README](https://github.com/cloudfoundry/metric-proxy/blob/master/README.md)

@hev 
@heycait 
@dtimm

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
+++ b/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
@@ -1,0 +1,18 @@
+#@data/values
+---
+system_domain: ""
+system_namespace: ""
+
+images:
+  metric_proxy: "oratos/metric-proxy"
+
+metric_proxy:
+  ca:
+    secret_name: "metric-proxy-ca"
+    crt: "" #! Base64-encoded ca for the metric-proxy-ca
+    key: "" #! Base64-encoded private key for the cert above
+  cert:
+    secret_name: "metric-proxy-cert"
+    crt: "" #! Base64-encoded cert for the metric-proxy server requires CN of metric-proxy
+    key: "" #! Base64-encoded private key for the cert above
+
--- a/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
+++ b/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
@@ -4,7 +4,7 @@ system_domain: ""
 system_namespace: ""
 
 images:
-  metric_proxy: "oratos/metric-proxy"
+  metric_proxy: "oratos/metric-proxy@sha256:a2a0d2201d1a57602a3db337bfa256d6e042dfc09a63ba1b6f39c952847e00dc"
 
 metric_proxy:
   ca:
--- a/config/metrics.yml
+++ b/config/metrics.yml
@@ -1,0 +1,24 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+#@ load("@ytt:base64", "base64")
+
+#@ def metrics_values():
+system_domain: #@ data.values.system_domain
+system_namespace: #@ data.values.system_namespace
+
+metric_proxy:
+  ca:
+    secret_name: "metric-proxy-ca"
+    crt: #@ data.values.metric_proxy.ca.crt
+    key: #@ data.values.metric_proxy.ca.key
+  cert:
+    secret_name: "metric-proxy-cert"
+    crt: #@ data.values.metric_proxy.cert.crt
+    key: #@ data.values.metric_proxy.cert.key
+#@ end
+
+#@ metrics = library.get("github.com/cloudfoundry/metric-proxy")
+--- #@ template.replace(metrics.with_data_values(metrics_values()).eval())
+
--- a/config/metrics.yml
+++ b/config/metrics.yml
@@ -1,8 +1,6 @@
 #@ load("@ytt:data", "data")
 #@ load("@ytt:library", "library")
-#@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:template", "template")
-#@ load("@ytt:base64", "base64")
 
 #@ def metrics_values():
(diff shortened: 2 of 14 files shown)


davewalter:
Hi @kkburr @heycait @dtimm,

Could one of you please rebase this against the latest changes on develop to resolve the conflict and allow our CI tests to run? We renamed the `kpack.registry` key right below your change in the values file and when we tried to rebase it, it messed up the authorship of your commit.

Thanks,
Dave and @acosta11

kkburr:
Dave & @acosta11 done! Thanks!

kkburr:
@acosta11 I think the pipeline failures are unrelated to us:

**check-vendir-sync:** running check-vendir-sync on our branch, I see the eirini vendir directory has a change. It reverts [this recent change](https://github.com/cloudfoundry/cf-for-k8s/commit/3ceff182e0385fdb174cb3e2e2f880701e0a0d3c#diff-5ccf46d2852f0f0e3689b5149d03c5f8R9) and fails the build

**tests:** this is the same failure I saw for my other PR (which is a very trivial change). The task "install-cf" is timing out on `bin/install-cf.sh cf-install-values.yml`

kkburr:
@jamespollard8 thanks, done!

kkburr:
We got the smoke tests to pass for us: https://loggregator.ci.cf-app.com/teams/main/pipelines/metric-proxy/jobs/run-tests-on-cf-for-k8s-pr/builds/45

acosta11:
> We got the smoke tests to pass for us: https://loggregator.ci.cf-app.com/teams/main/pipelines/metric-proxy/jobs/run-tests-on-cf-for-k8s-pr/builds/45

Re-ran the CI tests and it looks good! Thank you for the updates and sorry about the technical difficulties on our end!

Birdrock:
Hi @kkburr - looking at previous test runs, we see metrics present (with all zeros). Do you expect this PR to surface real metrics, or will it still display zero values for cpu, memory, and disk?

hev:
@Birdrock - it should return real metrics. Do you see metrics when using `kubctl top pods`?

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
```diff
@@ -0,0 +1,18 @@
+#@data/values
+---
+system_domain: ""
+system_namespace: ""
+
+images:
+  metric_proxy: "oratos/metric-proxy"
```
zankich: @kkburr Can you pin this image to the specific digest you'd like to include with this PR?

zankich:
Does this PR mean that this value https://github.com/cloudfoundry/cf-for-k8s/blob/ada64d42c467cc96d64e984c16154f79c87d3523/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/ccng-config.lib.yml#L86 can be set to `false`?

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
kkburr: Sure, we can do that if it's required. 👍

kkburr:
> Does this PR mean that this value
> 
> https://github.com/cloudfoundry/cf-for-k8s/blob/ada64d42c467cc96d64e984c16154f79c87d3523/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/ccng-config.lib.yml#L86
> 
> can be set to `false`?

@zankich I guess in theory yes. It feels like there's no harm in leaving it as is though.

Birdrock:
@kkburr I verified that metrics are being picked up via a manual deploy to a gke cluster. Metrics take a short amount of time to register, so they were not showing in our CI test runs. We will try flipping `temporary_ignore_server_unavailable_errors` to false for informational purposes. Changing this value would require a change to capi-k8s-release, which seems to be outside of the scope of this PR.

Birdrock:
We tried deploying with the flag set to false. The push still completely successfully, but we noted that it was slow to report success (~45 seconds on a very simple application). We do not have evidence that points to any specific cause, so this is more of a footnote.

config/metrics.yml
```diff
@@ -0,0 +1,24 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:library", "library")
+#@ load("@ytt:overlay", "overlay")
```
Birdrock, line 3: Are overlay or base64 (line 5) used?

tests/smoke/smoke_test.go
```diff
@@ -108,6 +108,12 @@ var _ = Describe("Smoke Tests", func() {
 				cfLogs := cf.Cf("logs", appName, "--recent")
 				return string(cfLogs.Wait().Out.Contents())
 			}, 2*time.Minute, 2*time.Second).Should(ContainSubstring("Hello World from index"))
+
+			By("verifying that the application's metrics are available.")
+			Eventually(func() string {
+				cfApp := cf.Cf("app", appName)
+				return string(cfApp.Wait().Out.Contents())
+			}).Should(ContainSubstring("cpu"))
```
Birdrock, line 116: Won't the string "cpu" always present, regardless of whether metrics are being reported?

jenspinney: Update: We don't think it is worth blocking the PR merge on this, but we're planning to remove this assertion after merging. The smoke tests are very basic (just testing you can push, curl, and get logs for an app), and this assertion feels a little out of place in that. Plus, it seems like it wouldn't likely catch any problems because of what @Birdrock mentioned.

jenspinney:
Thanks for submitting this (and for being responsive through all the back-and-forth!). Looks good 👍

PR #94, opened 2020-03-26
Support external-dns and static ip for ingress

paulczar:

Adds support for using a static ip for ingress, useful to avoid waiting for DNS updates.

Adds support to let external-dns manage DNS.

---

_Special notes for your reviewer /  **Acceptance Criteria** (if applicable)_

> For a PR that contains a new Deployment resource definition
**When** I run `kubectl get svc -n istio-system istio-ingressgateway -o jsonpath='{.status.loadBalancer.ingress[*].ip}'`
**Then** I see that the IP matches the IP I set in `ingress_static_ip`.

> For a PR that contains a new Deployment resource definition
**Given** I've deployed cf-for-k8s with this branch on a cluster than has external-dns controller
**Then** I see that my DNS is managed for me.


config-optional/use-external-dns-for-wildcard.yml
```diff
@@ -0,0 +1,13 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:template", "template")
+
+#@overlay/match by=overlay.subset({"kind": "Service", "metadata":{"name": "istio-ingressgateway"}})
+---
+metadata:
+  #@overlay/replace
```
cppforlife: are we intentionally replacing entire annotations map with below two?

config/values.yml
```diff
@@ -1,5 +1,6 @@
 #@data/values
 ---
+
```
cppforlife: extra line

sample-cf-install-values.yml
```diff
@@ -5,6 +5,9 @@ app_domains:
 #@overlay/append
 - apps.cf.example.com
 
+#! set this if you want to use a reserved static IP for istio's load balancer.
```
cppforlife: dont think we should add this to sample, as i think we were trying to keep sample-cf-install-values.yml to only inlcude required values.

config-optional/use-external-dns-for-wildcard.yml
paulczar: the existing [annotation](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml#L6148) is `metadata.annotations: null`  so we can't append to it.   It would need to be fixed to `metadata.annotations: {}` or similar to allow me to simply append them.   Since it tells us not to edit that generated yaml ... I have to brute force it like this.

nthomson-pivotal:
So I've tried this with the external-dns Istio integration that you can opt-in to. It still needs the FQDN change but doesn't need the extra annotations on the `Service`. My initial thoughts are that this seems like a nicer approach since it doesn't introduce external-dns annotations.

paulczar:
@nthomson-pivotal the reason I went this route, is that it doesn't require an extra arg for external-dns  to enable supporting ingress (`--source=istio-gateway`) which arguably isn't a big deal, but either way, adding the annotations makes it a little less work for the end user as it just works with a default external-dns runner.  Happy to do it either way though.

jamespollard8:
Hey @paulczar, `vendir` was out of sync on develop but we've fixed that. If you could rebase off (or merge in) new HEAD of develop, that should fix the failing check. Sorry about that!

config-optional/use-external-dns-for-wildcard.yml
cppforlife: i see. this is what we've done somewhere else for now as an additional overlay before "real" one:

```yaml
#! Currently ytt does not allow overlaying maps on top of nils;
#! remove nil annotations before further overlays (https://github.com/k14s/ytt/issues/108)
#@overlay/match by=overlay.subset({"metadata":{"annotations": None}}),expects="1+"
---
metadata:
  #@overlay/remove
  annotations:
```

replacing all annotations on this resource may have adverse effects.

paulczar: Should we at that `remove` in somewhere so I can dump the `replace`?   or ask upstream to change it from `nil` to `{}` ?

cppforlife: if above is included in cf-for-k8s repo it will be ok too.

paulczar: Added the above as an extra overlay in `config/` and removed my `replace`

paulczar:
What's stopping this from being merged? I don't want to have to keep managing merge conficts in the deploy docs.

jamespollard8:
> What's stopping this from being merged? I don't want to have to keep managing merge conficts in the deploy docs.

Sorry about that Paul. Our process relies on you (the PR author)(or anybody) posting an update via github message on the PR to trigger "new activity" to get our eyes back on the PR. I should have mentioned that in my previous message.

Our community pair should be able to get to this after lunch today.

PR #95, opened 2020-03-26
Change update-gcp-dns.sh to exclude 'Address' with # in addition to :

kkburr:
When we run `nslookup "*.$DNS_DOMAIN"`, our output includes the following line:

`Address:	10.35.30.10#53`

This causes the script to hang forever, since it should ignore that Address line.

This change excludes that line.

### DIFF
--- a/hack/update-gcp-dns.sh
+++ b/hack/update-gcp-dns.sh
@@ -40,7 +40,7 @@ resolved_ip=''
 while [ "$resolved_ip" != "$external_static_ip" ]; do
   echo "Waiting for DNS to propagate..."
   sleep 5
-  resolved_ip=$(nslookup "*.$DNS_DOMAIN" | grep Address | grep -v ':53' | cut -d ' ' -f2)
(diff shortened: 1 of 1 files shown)


davewalter:
Hi @kkburr,

Could you please update the description of this PR with some details around the use case for this change?

Thanks,
Dave and @acosta11

kkburr:
@acosta11 See the change to the description.

Also, it looks like the pipeline is failing to build in the install-cf task, which should be unrelated to this change.

jamespollard8:
Hey @kkburr , again you'll need to rebase off (or merge in) new HEAD of develop to fix the failing check.

jamespollard8:
If you have one handy, could you please show an example `nslookup` output that you're seeing? (before the `grep -v`s)

kkburr:
> If you have one handy, could you please show an example `nslookup` output that you're seeing? (before the `grep -v`s)

```
nslookup "*.denver-metric-proxy-pipeline.loggr.cf-app.com"
Server:		10.55.2.10
Address:	10.55.2.10#53

Non-authoritative answer:
Name:	*.denver-metric-proxy-pipeline.loggr.cf-app.com
Address: 35.238.251.241
```

jamespollard8:
Gotcha - thank you!

PR #105, opened 2020-03-30
Update deploy/kind/cluster.yml

julian-hj:
cf-for-k8s now deploys on k8s versions >= 1.16, so we no longer need to pin to k8s 1.15

**Acceptance Steps**
- run the steps to [deploy on kind](https://github.com/cloudfoundry/cf-for-k8s/blob/master/docs/deploy-local.md#steps-to-deploy-on-kind)
- run `kubectl version` and validate that the server version is > 1.16

@cloudfoundry/cf-volume-services

### DIFF
--- a/deploy/kind/cluster.yml
+++ b/deploy/kind/cluster.yml
@@ -2,7 +2,6 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
 - role: control-plane
-  image: kindest/node:v1.15.7
   kubeadmConfigPatches:
   - |
     kind: InitConfiguration


pivotaljohn:
`check-vendir-sync` check run was a failure from an upstream PR.  Cycling this PR to trigger rechecking.

jamespollard8:
Hey @julian-hj, thanks for the PR. `vendir` was out of sync on develop but we've fixed that. If you could rebase off (or merge in) new HEAD of develop, that should fix the failing check. Sorry about that!

jamespollard8:
Actually, the change is straight-forward and valuable so we'll go ahead and :shipit: as-is.

PR #117, opened 2020-04-03
Updates to the main deploy doc

Syerram:
### Updates
- Remove the friction of trying out cf-for-k8s by promoting docker hub over GCR. Most users who work with containers will have a dockerhub account than a GCR account (plus, it's a lot easier to signup for dockerhub).
- Promote source code based app over docker based app. The entire premise of cf-push is 12-factor apps. From onsi,
```
cf push :         Here is my source code
                        Run it on the cloud for me
                        I do not care how ...
```
- Move instructions to configure registries in one place rather than in multiple places.
- The docs are missing a few steps
- Move notes to quote block, so they don't distract from the instructions.

**Acceptance Steps**
As a user, 
When I follow the steps as is
Then I am able to successfully install CF and validate cf push.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -1,15 +1,14 @@
-# Deploying CF for K8s
+# Deploying Cloud Foundry on a Kubernetes cluster
 
 - [Prerequisites](#prerequisites)
   * [Required Tools](#required-tools)
   * [Kubernetes Cluster Requirements](#kubernetes-cluster-requirements)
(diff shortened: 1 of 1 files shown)


acosta11:
Manually updated table of content broken links, whitespace, and consistency of "cf-for-k8s", then merged in https://github.com/cloudfoundry/cf-for-k8s/commit/3b886eaf736387e87248f9ab401a8014272011a1 .

PR #84, opened 2020-03-24
Update cf-k8s-networking: IngressGateway Daemonset & NetworkPolicy Fixes

XanderStrike:
**Please describe this change**

- now deploys one ingressgateway per node [#171067809](https://www.pivotaltracker.com/story/show/171067809)
  - in demos, users can now hit any one of the nodes to get the full
  ingressgateway experience
  - in production, operators can set up hardware load balancers to
  balance between node ips

also featuring:
- minor updates to security policies to ensure kube-api-server has
  network access to the istio-sidecar-injector. See:
  https://github.com/cloudfoundry/cf-for-k8s/issues/69 and
  [#171877759](https://www.pivotaltracker.com/story/show/171877759)

**Acceptance Steps**

**GIVEN** I have deployed a cloud foundry
**WHEN** I `curl -H host:api.<cf-domain> <node ip>`
**THEN** I see some delicious json from capi

**GIVEN** I have deployed an app to my cloud foundry
**WHEN** I `curl -H host:<app fqdn> <node ip>`
**THEN** I see the response I expect from my app

@tcdowney @ndhanushkodi @rosenhouse @kauana @christianang 

Requestors on the issue: @kramerul @jkbschmid @emalm

Issue #123, opened 2020-04-07
CF-StatsUnavailable causing cf push failures when cluster does not include a metrics-server


cmoulliard:
## Issue

The CF client is getting the following error from I suppose the CAPI when we try to get the status of the application

```
cf app "test-app" -v
...
X-Xss-Protection: 1; mode=block
{   "errors": [;     {       "code": 200002,;       "detail": "Stats unavailable: Stats server temporarily unavailable.",;       "title": "CF-StatsUnavailable";     };   ]; }
Stats unavailable: Stats server temporarily unavailable.
```

Full log of the request : https://gist.github.com/cmoulliard/e73ab47d247c3010b974f9492928f175

## How to reproduce

- Install cf-for-k8s master branch on a k8s cluster
- Execute the following instructions to create an application
```
cf api --skip-ssl-validation https://api.${IP}.nip.io
cat /tmp/cf-values.yml | grep admin
cf auth xyge3gnqkme187rulw48
cf auth admin xyge3gnqkme187rulw48
cf create-org redhat.com
cf create-space -o redhat.com demo
cf target -o redhat.com -s demo
cf push test-app -o cloudfoundry/diego-docker-app
```

## Additional information

Log of the capi server is reporting such error
```
{"timestamp":1586251304.9358523,"message":"Statsd: cc.requests.outstanding:-1|c\ncc.requests.completed:1|c\ncc.http_status.2XX:1|c","log_level":"debug","source":"statsd.client","data":{},"thread_id":47059526067820,"fiber_id":47059521024500,"process_id":1,"file":"/usr/local/lib/ruby/gems/2.5.0/gems/statsd-ruby-1.4.0/lib/statsd.rb","lineno":457,"method":"send_to_socket"}
{"timestamp":1586251304.9360225,"message":"Statsd: Errno::ECONNREFUSED Connection refused","log_level":"error","source":"statsd.client","data":{},"thread_id":47059526067820,"fiber_id":47059521024500,"process_id":1,"file":"/usr/local/lib/ruby/gems/2.5.0/gems/statsd-ruby-1.4.0/lib/statsd.rb","lineno":477,"method":"rescue in send_to_socket"}
[...]

PR #136, opened 2020-04-09
updates for kind users

zankich:
Since https://github.com/cloudfoundry/cf-for-k8s/pull/84 has been merged, we no longer need to specify the NodePort. 

Added remove-ingressgateway-service.yml because kind doesn't support
LoadBalancer services

---

**Acceptance Steps**


jamespollard8:
This worked well for me using Kind on K8s v1.16.4 - thanks for the update!

As @zankich has pointed out, for Kind, we currently need to install a metrics-server to get smoke tests to pass. It's already captured by https://github.com/cloudfoundry/cf-for-k8s/issues/123

I'll add a note about that now to our documentation.

PR #143, opened 2020-04-15
Add documentation how to enable external ccdb und uaadb

c0d1ngm0nk3y:
I think it might be helpful to add some documentation to configure the installed cloud foundry in a certain way. At some point, there will probably be a more general approach to document stuff like this, but in the meantime this would have helped us to apply the wanted configuration.

Please have a look if you see this as beneficial as well.

### DIFF
--- a/docs/configuration/external-databases.md
+++ b/docs/configuration/external-databases.md
@@ -1,18 +1,52 @@
 # Using external databases
 
-## Cloud controller database
 
-You can use an external database for the cloud controller by providing following values:
+You can use an external database for the cloud controller and uaa by providing following values:
 
 ```yaml
 #@data/values
 ---
 capi:
   database:
-    adapter: <postgres | mysql2>
+    adapter: postgres
+    host: <host>
+    port: <port>
+    user: <user>
+    password: <password>
+    name: <database>
+
+uaa:
+  database:
+    adapter: postgresql
     host: <host>
     port: <port>
     user: <user>
     password: <password>
     name: <database>
 ```
+
+
+Please note to the **different spelling** of `postgres`  and `postgresql` for capi and uaa.
+
+As prerequisite, you need to execute the following step inside your postgres installation
+
+```bash
+VALUES_FILE=<path to file from above>
+VALUES_JSON=$(yaml2json "$VALUES_FILE")
+CCDB_USERNAME=$(jq -r '.capi.database.user' <<< "$VALUES_JSON")
+CCDB_PASSWORD=$(jq -r '.capi.database.password' <<< "$VALUES_JSON")
+CCDB_NAME=$(jq -r '.capi.database.name' <<< "$VALUES_JSON")
+UAADB_USERNAME=$(jq -r '.uaa.database.user' <<< "$VALUES_JSON")
+UAADB_PASSWORD=$(jq -r '.uaa.database.password' <<< "$VALUES_JSON")
+UAADB_NAME=$(jq -r '.uaa.database.name' <<< "$VALUES_JSON")
+cat > /tmp/setup_db.sql <<EOT
+CREATE DATABASE ${CCDB_NAME};
+CREATE ROLE ${CCDB_USERNAME} LOGIN PASSWORD '${CCDB_PASSWORD}';
+CREATE DATABASE ${UAADB_NAME};
+CREATE ROLE ${UAADB_USERNAME} LOGIN PASSWORD '${UAADB_PASSWORD}';
+EOT
+psql -U postgres -f /tmp/setup_db.sql
+psql -U postgres -d "${CCDB_NAME}" -c "CREATE EXTENSION citext"
+psql -U postgres -d "${UAADB_NAME}" -c "CREATE EXTENSION citext"
+```
+
--- a/docs/configuration/external-databases.md
+++ b/docs/configuration/external-databases.md
@@ -28,25 +28,101 @@ uaa:
 
 Please note to the **different spelling** of `postgres`  and `postgresql` for capi and uaa.
 
-As prerequisite, you need to execute the following step inside your postgres installation
+As prerequisite, you need to execute the following steps to configure your postgres installation for cf-for-k8s:
 
-```bash
-VALUES_FILE=<path to file from above>
-VALUES_JSON=$(yaml2json "$VALUES_FILE")
-CCDB_USERNAME=$(jq -r '.capi.database.user' <<< "$VALUES_JSON")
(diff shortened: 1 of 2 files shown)


docs/configuration/external-databases.md
```diff
@@ -0,0 +1,18 @@
+# Using external databases
+
+## Cloud controller database
+
+You can use an external database for the cloud controller by applying an overlay like
```
cppforlife, line 5: terminology nit pick: this is not an overlay. below examples just sets data values.

c0d1ngm0nk3y: That is of course right. You would probably want to apply those values as an overlay. I will change it.

Syerram: Hello Ralf,

Thanks for your contribution. Have you tested this with an external database? does it work? 

IMO, this should be in  `sample-cf-install-values.yml` as an option (the yml should be commented and have instruction for user).

loewenstein: I am not sure I agree to move this to `sample-cf-install-values.yml`, because a YAML file is not really readable documentation and this particular configuration is optional.

jamespollard8: > Hello Ralf,
> 
> Thanks for your contribution. Have you tested this with an external database? does it work?
> 
> IMO, this should be in `sample-cf-install-values.yml` as an option (the yml should be commented and have instruction for user).

hey @c0d1ngm0nk3y, our PM is interested in hearing if you've tested this already.

drnic:
Either in this PR, or a future one, we can also mention that there are similar but diff `data.values.uaa.database` to provide for the UAA.

docs/configuration/external-databases.md
c0d1ngm0nk3y: @jamespollard8 We did of course test it before that PR and it worked as expected. But we faced some problems when updating our cf-for-k8s. Not sure if this was really related, but we decided to revert using the external ccdb for now.

c0d1ngm0nk3y:
> Either in this PR, or a future one, we can also mention that there are similar but diff `data.values.uaa.database` to provide for the UAA.
> 
> https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/uaa.yml#L8-L16

We had this planned too. But first we wanted to actually use an external database for uaa and see how this PR works out.

Syerram:
@c0d1ngm0nk3y AFIAK, this will not work on an external database. We seed the database on the postgres pod using a [script](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/postgres.yml#L65) that is passed to the local postgres pod property `custom-init-script`. 

With an external database, this will not be run and you will receive an error during install (unless you run the script yourself or have a pre-seeded database). We have a roadmap item to enable an external database and run seed scripts and migrations. The idea is to use the same interface for the internal database that comes default with cf-for-k8s.

loewenstein:
@Syerram That indeed looks problematic. Why does the script have to run on the Postgres Pod? How is that done in cf4bosh?

Syerram:
@loewenstein In cf-4-bosh, it is an ops files and it expects users to create the database/username/passwords before they can connect CF to the external database. In the first iteration of supporting external DBs, we will follow the same experience i.e. user will create the databases, uid/passwords . We will provide documentation for users so that they can correctly seed the external database.

loewenstein:
@syerram so is adding documentation on how to correctly seed the database the missing part to make this PR merge ready?

If not, what else should we provide?

Syerram:
As I mentioned, we have a explore story [1] in our backlog that will investigate how-to configure an external database. We will add this PR to that "explore story". This PR is one way to define the external DB interface (including where the config lives). 

We also want to know what additional things need to be true for users to configure an external database - for example, we don't want to install the Postgres DB if the user provides an external database. Right now, we don't have that capability.

[1] https://www.pivotaltracker.com/story/show/172561146

kramerul:
I tested the external database configuration also with UAA.

paulczar:
Here's an example set of overlays I used to do this on GKE using CloudSQL and a SQL proxy.  The SQL proxy has the same service name as the built-in postgres, so the config for CC and UAA do not change.

https://github.com/paulczar/cf-for-gke

c0d1ngm0nk3y:
What is the current state? Is there still something missing in this PR?

paulcwarren:
This issue is linked to the PR to actually add support for external databases and [that story](https://www.pivotaltracker.com/story/show/173372976) is currently being discussed by the team.

docs/configuration/external-databases.md
```diff
@@ -0,0 +1,138 @@
...
+
+1. Create configuration file `db-values.yaml`
+
+    ```yaml
+    #@data/values
+    ---
+    capi:
+      database:
+        adapter: postgres
+        host: postgres-postgresql.default.svc.cluster.local
```
Birdrock, line 95: Kubernetes could not find a service with this host. Following these instructions, the UAA and API server failed due to the incorrect service name.

We were able to get it to work by changing the host to the provided value from the Helm installation. The Helm chart installs a Postgres service with host of `postgresql.default.svc.cluster.local`.

docs/configuration/external-databases.md
```diff
@@ -0,0 +1,138 @@
...
+        port: 5432
+        user: capi_user
+        password: capi_password
+        name: capi_db
+
+    uaa:
+      database:
+        adapter: postgresql
+        host: postgres-postgresql.default.svc.cluster.local
```
Birdrock, line 104: See comment above.

docs/configuration/external-databases.md
```diff
@@ -0,0 +1,138 @@
...
+    export PGHOST=<host where postres is running>
+    ```
+
+2. Run the following script. It will
+   * create one database each for Cloud Controller and UAA
+   * create one user each for Cloud Controller and UAA
+   * activate the `citext` extension for each of these databases
+
+    ```bash
+    VALUES_JSON=$(yaml2json "$VALUES_FILE")
```
Birdrock, line 57: There doesn't seem to be a clear utility with the name of `yaml2json`. I found a few different projects, but was unsure which one was used. Can we remove the dependency on `yaml2json`?

Alternatives:
* Use [yq](https://github.com/mikefarah/yq).
* Have the shell script export values and interpolate them into a heredoc that creates the db-values.yaml.

kramerul: I will change it to yq

docs/configuration/external-databases.md
kramerul: That's correct. I will change it.

Birdrock:
The messaging for the above message is a bit off - the uptimer results were appended rather than added as a separate message.

Thanks for making the requested changes!

PR #146, opened 2020-04-16
[docs] recommend a separate user-provided app-registry.yml, instead of modifying cf-values.yml

drnic:
Instead of asking users to modify the generated `/tmp/cf-values.yml` file, the documentation now asks them to create a `/tmp/app-registry.yml` that is separate from `cf-values.yml`. This allows a new user to generate cf-values.yml multiple times whilst keeping their static app-registry.yml file.

The `bin/install-cf.sh` script has been upgraded to allow an arbitrary number of `ytt -f file1.yml -f file2.yml` files, via `bin/install-cf.sh file1.yml file2.yml`

Update: I also used this approach to specify a static IP for the ingress, by creating a `/tmp/static-up.yml`:

```yaml
#@data/values
---
istio_static_ip: 34.83.153.141
```

### DIFF
--- a/bin/install-cf.sh
+++ b/bin/install-cf.sh
@@ -10,12 +10,15 @@ fi
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null 2>&1 && pwd)"
 CONFIG_DIR="${SCRIPT_DIR}/../config"
 
-cf_install_values_path="$1"
+ytt_args=()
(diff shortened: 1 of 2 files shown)


jamespollard8:
Thanks Nic, our PM @Syerram will review this when he gets a chance next week.

jamespollard8:
We've removed the install-cf script which should make using the strategy you've suggested more intuitive 👍

PR #147, opened 2020-04-16
Ignore tmp

drnic:
Currently the docs ask you to create files in `/tmp/...` but it is also common for people to assume they can use a local `tmp/` inside the project. This PR adds `tmp` to the gitignore.


jamespollard8:
Done - thanks Doc!
2780b2009cd7d635781287a8dfd671ce23ef4781

Issue #102, opened 2020-03-30
xxx-generated-istio-yaml has incorrect DaemonSet spec


kramerul:
If you look at [this line](https://github.com/cloudfoundry/cf-for-k8s/blob/d6252e2c89e2224742a0c55a40573866332ae3c9/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml#L5853) and compare it to the API of [DaemonSet](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.16/#daemonsetspec-v1-apps) you can see that `strategy` is no valid spec element. It should be `updateStrategy`

Nether the less it seems that cf-for-k8s can be installed without error.

jamespollard8:
Great catch - thanks for reporting it. @cloudfoundry/cf-networking could you please update to `updateStrategy`? Also, according to the docs, the `maxSurge` property might no longer be available.

kramerul:
Could you please also enhance kapp, that it reports such Problems. It seems that kapp is ignoring unknown fields in the spec.

jamespollard8:
> Could you please also enhance kapp, that it reports such Problems. It seems that kapp is ignoring unknown fields in the spec.

@cppforlife would it make sense to add such logic into `kapp`?

cppforlife:
> Could you please also enhance kapp, that it reports such Problems. It seems that kapp is ignoring unknown fields in the spec.

kapp doesnt do anything about contents of the resource. it strictly relies on k8s api to give your errors with resources. im not sure why k8s api doesnt throw some error about unkown field...

kauana:
> Great catch - thanks for reporting it. @cloudfoundry/cf-networking could you please update to `updateStrategy`? Also, according to the docs, the `maxSurge` property might no longer be available.

@jamespollard8 thanks for letting us know. It seems that DaemonSets doesn't have a `maxSurge` field after all so we will update that as well.

acosta11:
Fixed by #124 which is now pushed to master in c23c954 .

Thanks,
Andrew and @ericpromislow (Community Pair)

PR #156, opened 2020-04-23
Add validation with kubeval

jkbschmid:
Add a `--validate` flag to install-cf.sh so we can validate k8s resources.
Use `--validate` to test PRs.

Solves https://github.com/k14s/kapp/issues/98  

~This is WIP because kapp's `Config` resource does not play well with kubeval.~

**Acceptance Steps**

Introduce an error somewhere in the `config` folder and execute `install-cf.sh --validate` to see if it is caught.

### DIFF
--- a/config/kapp-rebase-rules.yml
+++ b/config/kapp-rebase-rules.yml
@@ -2,8 +2,7 @@
 apiVersion: kapp.k14s.io/v1alpha1
 kind: Config
 metadata:
-  name: "test"
-  namespace: "test"
+  name: "kapp-config"
 rebaseRules:
 - path: [metadata, annotations, pv.kubernetes.io/bind-completed]
   type: copy
--- a/config/kapp-rebase-rules.yml
+++ b/config/kapp-rebase-rules.yml
@@ -1,7 +1,9 @@
 ---
 apiVersion: kapp.k14s.io/v1alpha1
 kind: Config
-
+metadata:
+  name: "test"
+  namespace: "test"
 rebaseRules:
 - path: [metadata, annotations, pv.kubernetes.io/bind-completed]
   type: copy
--- a/tests/configs/configs_test.go
+++ b/tests/configs/configs_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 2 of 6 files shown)


jkbschmid:
For the validation to pass, we'd also have to bump metric-proxy in the vendir.yml:
```
url: https://github.com/cloudfoundry/metric-proxy
-      ref: 614dbcd1a80512fb707f936067d8c498e3471682
+      ref: 8cda308b4681174df273ccb9cf9a7fdebdc7b053
```
Is this feasible, @heycait?

heycait:
Yes it should be fine to bump to that sha.

davewalter:
Hi @jkbschmid,

Thanks for the contribution. Ironically, we are actually trying to eliminate the need for the `install-cf.sh` script completely, and so are reluctant to add more logic to it. We feel that the issues you are encountering with regards to need to validate the rendered templates before deploying them is actually a problem with `kapp`, as opposed to `cf-for-k8s` itself. Is there any reason why you cannot simply run this validation yourself before running `install-cf.sh`?

Regards,
Dave and @Birdrock

jkbschmid:
Hi @davewalter,
> Is there any reason why you cannot simply run this validation yourself before running install-cf.sh?

Of course, I can run the validation myself, but that will only prevent _me_ from contributing erroneous k8s objects.^^ If kapp does not take on this validation, then it's very easy for people to get e.g. an indentation wrong and not notice. #102 and cloudfoundry/metric-proxy#2 show that this issue is not just theoretical. Therefore, I think it would be very valuable to have some kind of PR validation that checks whether k8s objects are actually well-formed, do you agree?

> We feel that the issues you are encountering with regards to need to validate the rendered templates before deploying them is actually a problem with kapp

I completely agree with you that this should be handled in kapp as, normally, kubectl would perform this kind of validation. This was intended as a quick solution to prevent any further bugs, since it is not sure if or when kapp will support this. 

Maybe instead of adding to the install script, we could add a PR-job that validates k8s objects, what do you think?

jamespollard8:
`install-cf.sh` is now officially gone 😁 

>Maybe we could add a PR-job that validates k8s objects, what do you think?

Yeah that's a good idea. But on every PR we're already doing more than just validating the objects; we're doing a full on deploy and run of smoketests which should give us ample coverage.

We did just add a unit test though. Maybe that would be a good place to add this validation?

Also, with a minimal amount of effort, I wasn't able to get `kubeval` passing on our current templated file. Are you currently able to get it to work?

```
$ ytt -f config -f tmp/values.yml > tmp/play.yml
$ kubeval -v 1.17.0 --strict --ignore-missing-schemas --skip-kinds Config ./tmp/play.yml
WARN - Set to ignore missing schemas
ERR  - ./tmp/play.yml: Missing 'metadata' key
```

jkbschmid:
> Maybe that would be a good place to add this validation?

Sure, I added a test that does this validation.

> Are you currently able to get it to work?

You need to add name/namespace to [kapp-rebase-rules.yml](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/kapp-rebase-rules.yml). I also bumped metric-proxy to make test pass.

jamespollard8:
Excellent - thanks! We're discussing as a team about adding this feature. We'll plan on making a decision early next week.

Either way, we'd like to handle the metrics-proxy bump as a separate PR.

tests/configs/configs_test.go
```diff
@@ -85,4 +89,22 @@ var _ = Describe("Optional Configs", func() {
 			Expect(count).To(BeNumerically(">=", 1))
 		})
 	})
+	When("validating with kubeval", func() {
+		It("should pass", func() {
+			command := exec.Command("kubeval", "--strict", "--ignore-missing-schemas", "--skip-kinds", "Config", templatedPath)
```
jamespollard8, line 94: Let's use https://github.com/cloudfoundry/cf-for-k8s/blob/master/supported_k8s_versions.yml to test kubeval with both the newest and oldest versions of K8s that we support

vendir.yml
```diff
@@ -30,7 +30,7 @@ directories:
   - path: github.com/cloudfoundry/metric-proxy
     git:
       url: https://github.com/cloudfoundry/metric-proxy
-      ref: 614dbcd1a80512fb707f936067d8c498e3471682
+      ref: 8cda308b4681174df273ccb9cf9a7fdebdc7b053
```
jamespollard8, line 33: We've bumped metric-proxy on develop. You should be able to rebase now to just apply the kubeval-related changes in this PR

config/kapp-rebase-rules.yml
```diff
@@ -1,7 +1,9 @@
 ---
 apiVersion: kapp.k14s.io/v1alpha1
 kind: Config
-
+metadata:
+  name: "test"
```
jamespollard8, line 5: maybe call this kapp-config? We'll also need an entry in the new file: https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/0-min-kapp-version.yml

config/kapp-rebase-rules.yml
```diff
@@ -1,7 +1,9 @@
...
+  namespace: "test"
```
jamespollard8, line 6: it seems that the namespace entry is not necessary

tests/configs/configs_test.go
```diff
@@ -85,4 +90,44 @@ var _ = Describe("Optional Configs", func() {
 			Expect(count).To(BeNumerically(">=", 1))
 		})
 	})
+	When("validating with kubeval", func() {
+		It("should pass", func() {
+			for _, v := range getSupportedK8Versions() {
+				By(fmt.Sprintf("checking version '%s'", v))
```
jamespollard8, line 96: maybe "checking with K8s version.." would be more clear?

tests/configs/configs_test.go
```diff
@@ -85,4 +90,44 @@ var _ = Describe("Optional Configs", func() {
...
+				print("kubeval", "--strict", "--ignore-missing-schemas", "--skip-kinds", "Config", "-v", v, templatedPath)
```
jamespollard8, line 97: When I run this I see some missing whitespace (spaces and newlines)
```
STEP: checking version '1.18'
kubeval--strict--ignore-missing-schemas--skip-kindsConfig-v1.18/tmp/cf-for-k8s.ymlSTEP: checking version '1.14'
kubeval--strict--ignore-missing-schemas--skip-kindsConfig-v1.14/tmp/cf-for-k8s.yml
```

tests/configs/configs_test.go
```diff
@@ -85,4 +90,44 @@ var _ = Describe("Optional Configs", func() {
...
+				command := exec.Command("kubeval", "--strict", "--ignore-missing-schemas", "--skip-kinds", "Config", "-v", v, templatedPath)
```
jamespollard8, line 98: Now that the Configs have metadata, I believe we can remove the skip Configs option

jamespollard8:
We add `kubeval` to our docker image directly on develop to help get this green

tests/configs/configs_test.go
jkbschmid: good point

tests/configs/configs_test.go
jkbschmid: I changed to use `command.args` which also prevents inconsistencies between the actual command and what we print.

jkbschmid:
@jamespollard8 Thanks for the review, let's see how many bugs this will actually catch ;)

PR #151, opened 2020-04-17
Pin cfroutesync image digest in cf-k8s-networking

ndhanushkodi:
[#172087151](https://www.pivotaltracker.com/story/show/172087151)

Related to [this issue](https://github.com/cloudfoundry/cf-k8s-networking/issues/36), we now pin the cfroutesync image digest in cf-k8s-networking, so we can remove the override for the image url in cf-for-k8s values. 

**Acceptance Steps**
**GIVEN** I have deployed a cloud foundry
**WHEN** I describe the cfroutesync deployment with kubectl -n cf-system describe deployment/cfroutesync
**THEN** I see the container spec specifies `gcr.io/cf-networking-images/cf-k8s-networking/cfroutesync@sha256:c54be4b33f6a8a7a6dc24ee74cff468b3484411b64182d866a5be52cd67425a4`

_Tag your pair, your PM, and/or team_

@christianang 

> _It's helpful to tag a few other folks on your team or your team alias in case we need to follow up later._

PR #159, opened 2020-04-23
Include Route CRD

jenspinney:
This PR includes the Route CRD that will be referenced by the Cloud Controller and the future routecontroller component from the cf-k8s-networking team.

---

This PR is a follow-up PR to the bump made as part of #151 . Until that PR is merged, this diff will look messy and misleading. To see what we've added as part of just this PR before that is merged, click on [this link](https://github.com/cloudfoundry/cf-for-k8s/compare/pin-cfroutesync-digest...pr-include-route-crd).

See [this slack thread](https://cloudfoundry.slack.com/archives/CFX13JK7B/p1587675432112300) for more context on why this CRD is needed urgently. Basically, we won't be able to bump capi-k8s-release in cf-for-k8s until this CRD is available in cf-for-k8s. For broader context on the general routecontroller work, see this story: https://www.pivotaltracker.com/story/show/171826991.

This PR does not include any actual functional changes. It adds what is for the moment an unused CRD, but that foundation is needed for the route-controller work the Networking and CAPI teams are doing.

**Acceptance Steps**

`kubectl get crds` should return "routes.networking.cloudfoundry.org" in its list.

Thanks,
Jen & @sethboyles (CF-k8s Networking and CAPI)

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
+++ b/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
@@ -1,0 +1,129 @@
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.4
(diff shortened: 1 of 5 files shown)


davewalter:
Hi @jenspinney,

Could you please change the base branch to `develop`?

Thanks,
Dave

davewalter:
Thanks @jenspinney!

PR #161, opened 2020-04-24
Move metacontroller to the build dir and use kbld to pin image to a digest

davewalter:
>

Move metacontroller to the build directory as this allows us to run kbld to pin the image to the digest once at packaging time instead of at install time. 

**Acceptance Steps**

This PR does not introduce any material changes. Acceptance should be as simple as confirming that nothing is broken.


build/metacontroller/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+echo "generating Metacontroller resource definitions..."
+ytt -f "${SCRIPT_DIR}/_vendir/manifests" \
+    -f "${SCRIPT_DIR}/pin-image-to-digest.yml" > "${SCRIPT_DIR}/../../config/_ytt_lib/metacontroller/rendered.yml"
```
cppforlife, line 8: @davewalter how about just using get-kbld.io to resolve digests. `kbld -f _vendir/manifests > ...`

Birdrock: @cppforlife Hi Dmitriy - thanks for the input. We have created a [chore](https://www.pivotaltracker.com/story/show/172570472) to investigate usage of kbld in the future. For the purposes of this PR, we suggest leaving it as is for the following reasons:
1. It adds an additional dependency to the build chain that isn't currently used in this fashion.
2. There are other cases that might merit usage as well; We'd like to keep everything as consistent as possible.

jamespollard8:
Dave and I got a chance to sync-up in person. We agreed that there's not much value in changing this since we're not expecting changes to metacontroller.

> Could this be done through just an overlay with a hardcoded image sha?
@Birdrock our suspicion was correct. We're already have this overlay here: https://github.com/cloudfoundry/cf-for-k8s/blob/677a6456cd00ccac0f2f5e4f66fd483a6e3b49cb/config/networking.yml#L36

Issue #46, opened 2020-02-26
Apps not available via HTTPS


acosta11:
## Summary
HTTPS traffic does not reach apps running on cf-for-k8s.

## Repro Steps
1. Deploy cf-for-k8s and log in via the cf CLI
1. Enable feature flag 'diego_docker'
1. Target a space and push a docker app
1. Attempt to curl the route provided in the push output and observe it is only available over unencrypted http.
```
> cf push myapp -o cloudfoundry/diego-docker-app

> curl -i http://myapp.<app-domain>
HTTP/1.1 200 OK
date: Wed, 26 Feb 2020 00:21:36 GMT
content-length: 1
content-type: text/plain; charset=utf-8
x-envoy-upstream-service-time: 3
server: istio-envoy

0

> curl -i https://myapp.<app-domain> -k
HTTP/2 404
date: Wed, 26 Feb 2020 00:21:18 GMT
server: istio-envoy
```

## Context
Currently, there are two istio ingress gateways: one for the cf-system components and one for cf-workloads. The ingress gateway for cf-workloads namespace only configures a rule for unencrypted http traffic to port 80. Meanwhile the system component ingress gateway only listens for encrypted https traffic on port 443. So any encrypted https traffic to port 443 only matches the routing rule on the cf-system gateway and receives an istio 404 response. We expect to work with the cf-for-k8s networking team to resolve this issue.

loewenstein:
Has this been solved by #64?

XanderStrike:
#64 provided documentation for how users could enable TLS by providing their own certs. The story title was misleading, I've updated the PR to better reflect the work that was actually done.

The story to enable TLS for ingress by default is [here](https://www.pivotaltracker.com/n/projects/2407973/stories/171870086) and it is prioritized high in our backlog.

jamespollard8:
Closed by #179

PR #166, opened 2020-04-29
Enable app ingress TLS by default and fix Route CRD

XanderStrike:
* This bump of CF K8s Networking includes two changes.
  1. Removes the Status section from the Route CRD
      [#172569406](https://www.pivotaltracker.com/story/show/172569406).
  2. Changes which gateway is associated with the virtual services for
      app routes. This is part of the work needed to have app ingress routing
      support TLS by default.

* This also includes the changes to cf-for-k8s to enable app TLS by default
  1. New requirement for distinct app and system domains is documented
      in values files
  2. Fields added for new cert to values files
  3. generate-values now creates a 2nd cert for app ingress
      automatically

fixes cloudfoundry/cf-for-k8s#46

**Acceptance Steps**

```
GIVEN a fresh kubernetes cluster
WHEN I install cf-for-k8s
AND I deploy an application
THEN I can reach my app via https
```

THE SQUAD: @jenspinney @ndhanushkodi @tcdowney @christianang @kauana @jrussett


config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
```diff
@@ -1,3 +1,4 @@
+
```
jamespollard8, line 1: nit: is this blank line intentional?

tcdowney: Yeah, Kubebuilder adds that blank line to its generated CRDs for some reason (maybe not intentional on their end). See the example `CronJob` CRD:

https://github.com/kubernetes-sigs/kubebuilder/blob/d6e2b995831ce26228f2303e9d1b3a55acd3d225/docs/book/src/cronjob-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml#L1

We had manually copy pasted it to our top level `config` directly previously I think which is why the blank line shows up in the diff here. This time I copied the generated file entirely.

jenspinney:
> Hello friend, it looks like your pull request has failed one or more of our checks. Please take a look! 👀

The tests failed because **we've added a new restriction that the system domain and apps domain be different**. The cf-for-k8s pipeline assumed they were the same. I've pushed a commit to this branch that I hope will fix it, but this may take some trial-and-error to get right.

jenspinney:
Actually, the smoke test run here won't reconfigure the pipeline, so I expect this still to fail. 🤔

jenspinney:
I think we'll have to do this PR in two steps:

1. A PR to change the generate-values script & cf-for-k8s CI so that the system domain and apps domain are different. After merging, Rel-int reconfigures their CI with the pipeline changes.
1. This PR, with the changes to the Route CRD and the functionality needed to have app ingress TLS working by default.

jenspinney:
After discussing with folks on Rel-Int, closing this PR in favor of just PRing the Route CRD change as a standalone thing and investigating whether we can continue to use a single domain for apps and system components in the rel-int CI.

PR #169, opened 2020-04-29
Bump cf-k8s-networking to fix reconcile problem with Route CRD

jenspinney:
## Summary
This PR bumps the cf-k8s-networking component. The effect of this bump is to make the Route CRD no longer cause problems when updating an existing cf-for-k8s environment with `kapp`.

## Details
(Copied from tracker story [#172569406](https://www.pivotaltracker.com/story/show/172569406))
The `Route` CRD has an empty top level `status` section in its definition which causes lifecycle problems with `kapp`.

https://github.com/cloudfoundry/cf-k8s-networking/blob/63b8a02d2ec88e5b21479d92cfa266d28dd5cc67/routecontroller/config/crd/bases/networking.cloudfoundry.org_routes.yaml#L125-L130

It will deploy successfully initially, but on upgrades may have failures because `kapp` will try and reconcile the CRD to have an empty status.

```
kapp deploy -a test -f config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
Target cluster 'https://34.83.92.22' (nodes: gke-tim-2020-04-28-default-pool-57185594-9sff, 2+)

Changes

Namespace  Name                                Kind                      Conds.  Age  Op  Wait to    Rs       Ri
(cluster)  routes.networking.cloudfoundry.org  CustomResourceDefinition  0/0 t   1m   -   reconcile  ongoing  Condition Established is not set

Op:      0 create, 0 delete, 0 update, 1 noop
Wait to: 1 reconcile, 0 delete, 0 noop

Continue? [yN]:
```

A future version of the controller-gen tooling that Kubebuilder uses will no longer generate the status section here (see https://github.com/kubernetes-sigs/controller-tools/pull/433), but in the meantime we can manually remove it.

---

## Acceptance Steps

**Given** I've previously deployed cf-for-k8s
**And** I re-deploy cf-for-k8s
**Then** I should not see the routes.networking.cloudfoundry.org custom resource definition stuck in the reconcile stage.

Thanks,
Jen, @tcdowney, @ndhanushkodi, @kauana, @XanderStrike, @jrussett, @christianang

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
+++ b/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
@@ -1,3 +1,4 @@
+
 ---
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
@@ -121,9 +122,3 @@ spec:
   - name: v1alpha1
     served: true
     storage: true
-status:
-  acceptedNames:
-    kind: ""
-    plural: ""
-  conditions: []
-  storedVersions: []
--- a/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/values.yaml
+++ b/config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/values.yaml
@@ -7,7 +7,7 @@ systemNamespace: cf-system
 workloadsNamespace: cf-workloads
 
 cfroutesync:
-  image: gcr.io/cf-networking-images/cf-k8s-networking/cfroutesync@sha256:ce8f2571251afe12810a37628865233e389ade3c086757e50ed9c1b309bbfad3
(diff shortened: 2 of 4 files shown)


config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/crd/networking.cloudfoundry.org_routes.yaml
```diff
@@ -1,3 +1,4 @@
+
```
tcdowney, line 1: This whitespace was intentionally added. Kubebuilder adds that blank line to its generated CRDs for some reason (maybe not intentional on their end). See the example `CronJob` CRD:

https://github.com/kubernetes-sigs/kubebuilder/blob/d6e2b995831ce26228f2303e9d1b3a55acd3d225/docs/book/src/cronjob-tutorial/testdata/project/config/crd/bases/batch.tutorial.kubebuilder.io_cronjobs.yaml#L1

We had manually copy pasted it to our top level `config` directly previously I think which is why the blank line shows up in the diff here. This time the generated file was copied wholesale.

davewalter:
Thanks @jenspinney

PR #171, opened 2020-04-30
update logging to 0.2.0

Benjamintf1:
- Build logs now included in cf-push
- remove uaa client from log-cache
- add forwarder api to fluentd

**Acceptance Steps**

- When cf-pushing an app with build logs they should appear while pushing

@pianohacker @MasslessParticle @JSchuenke


davewalter:
Thanks for the PR @Benjamintf1!

PR #172, opened 2020-04-30
VirtualService for UAA routes subdomains to UAA

### LINKED ISSUE #170: HTTP requests are not routed to subdomains of UAA
### Summary

Ingress routing should be configured to send requests to any subdomain of `uaa.system-domain` and of `login.system-domain` to UAA. This is required to support UAA's identity zones feature, which is UAA's implementation of multi-tenancy.

For reference, here is how cf-deployment implemented this requirement: https://github.com/cloudfoundry/cf-deployment/blob/50cfcf63e94b678a932516c277ece755cf1a140d/cf-deployment.yml#L717-L721

### Reproduction Steps
Using cf-for-k8s v0.1.0 deployed on GKE with k8s version 1.14.10-gke.27

Here is an integration test which should pass, but does not:

```
#!/bin/bash

set -eu

CF_VALUES_FILE_PATH=$1
SYSTEM_DOMAIN=$2

UAA_ADMIN_CLIENT_SECRET=$(bosh int $CF_VALUES_FILE_PATH --path "/uaa/admin_client_secret")

uaac target --skip-ssl-validation "https://login.${SYSTEM_DOMAIN}"
uaac token client get admin -s $UAA_ADMIN_CLIENT_SECRET
uaac curl -k -X POST /identity-zones -d '{"subdomain": "test-subdomain", "name": "test-subdomain"}' -H 'Content-Type: application/json'

# Not using the subdomain will work fine.
curl -fkv "https://login.${SYSTEM_DOMAIN}/.well-known/openid-configuration"

# Using the subdomain...
# Actual behavior: this will return a 404. Desired behavior: it should work, and return some json.
if curl -fkv "https://test-subdomain.login.${SYSTEM_DOMAIN}/.well-known/openid-configuration"; then
  echo "TEST PASSED"
else
  echo "TEST FAILED"
  exit 1
fi
```

cfryanr:
Fixes #170. See full description there.

**Acceptance Steps**

See issue #170 for acceptance test.

### DIFF
--- a/config/uaa.yml
+++ b/config/uaa.yml
@@ -234,6 +234,8 @@ spec:
   hosts:
   - #@ "uaa." + data.values.system_domain
   - #@ "login." + data.values.system_domain
+  - #@ "*.uaa." + data.values.system_domain
+  - #@ "*.login." + data.values.system_domain
   gateways:
   - #@ data.values.system_namespace + "/istio-ingressgateway"
   http:


cfryanr:
Oops. meant to submit this PR to the develop branch. Clicked the wrong button.

PR #174, opened 2020-05-01
ENH: bump CAPI to current head of master

jamespollard8:
>

Sai mentioned potentially cutting a 0.2.0 release and we pointed out that CAPI hasn't been bumped in a couple weeks.

This includes:
- a fix for a regression/bug where org deletion got busted
- https://github.com/cloudfoundry/capi-k8s-release/pull/33 which completes [#171007126](https://www.pivotaltracker.com/story/show/171007126), making sure CAPI communication uses all internal DNS (it wasn't for loggregator)

---

**Acceptance Steps**

PR checks passing should be sufficient. After merging, it will be interesting to check-in on our failing AKS validation and see if increasing the retry count on the ccdb-migration job helps mitigate that issue. see https://github.com/cloudfoundry/capi-k8s-release/issues/35

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/service-accounts.yml
+++ b/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/service-accounts.yml
@@ -19,6 +19,26 @@ roleRef:
   name: cluster-admin
   apiGroup: rbac.authorization.k8s.io
 
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: cc-worker-service-account
+  namespace: #@ data.values.system_namespace
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: cc-worker-service-account-superuser
+subjects:
+- kind: ServiceAccount
+  name: cc-worker-service-account
+  namespace: #@ data.values.system_namespace
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
+
 #! kpack registry credentials
 ---
 apiVersion: v1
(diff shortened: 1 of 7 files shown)


jamespollard8:
Review from @jspawar:
> great thanks for doing the bump. The only other change there [worth mentioning] should actually fix some regression/bug where org deletion got busted. So should be good to go

config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/service-accounts.yml
```diff
@@ -19,6 +19,26 @@ roleRef:
...
+kind: ClusterRoleBinding
+metadata:
+  name: cc-worker-service-account-superuser
+subjects:
+- kind: ServiceAccount
+  name: cc-worker-service-account
+  namespace: #@ data.values.system_namespace
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
```
zankich, line 39: @jamespollard8 @selzoc does this service account really need cluster-admin?

jspawar: I don't think we've done an audit yet of enumerating all permissions that particular service account needs, which of course we should certainly do soon. Fwiw to the best of my knowledge, that deployment needs at least permissions for CRUD operations on kpack Image CRs and I think maybe also route CRs (specifically when handling jobs to delete orgs/everything under an org)

jspawar: The API server deployment's service account is also still a cluster admin

jspawar: Just found a chore in our backlog to address this: https://www.pivotaltracker.com/story/show/171735131

selzoc: @zankich nope! we will eventually prioritize this: https://www.pivotaltracker.com/story/show/171735131

Can you give us a sense of how urgent the need is to reduce the scope of permissions?

PR #176, opened 2020-05-01
Update deploy instructions

Syerram:
Now that we are not using install script, adding more clarity on the raw commands.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -36,20 +36,15 @@ To deploy cf-for-k8s as is, the cluster should:
 - be running Kubernetes version 1.14.x, 1.15.x, or 1.16.x
 - have a minimum of 5 nodes
 - have a minimum of 3 CPU, 7.5GB memory per node
-- have a metrics-server
-  - This may be as simple as running something like `kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml`
-
-### IaaS Requirements
-
-- Supports `LoadBalancer` services
-- Supports `metrics-server`
+- support `LoadBalancer` services
+- support `metrics-server`
   - Most IaaSes come with `metrics-server`, but if yours does not come with it or if you're using `kind`, then you may want to run something like 
   
   ```console
     kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml
   ```
-  
-- Defines a default StorageClass
+
+- defines a default StorageClass
   - requires [additional config on vSphere](https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/storageclass.html), for example
 
 ### Requirements for pushing source-code based apps to Cloud Foundry foundation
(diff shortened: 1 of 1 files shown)


docs/deploy.md
```diff
@@ -36,20 +36,15 @@ To deploy cf-for-k8s as is, the cluster should:
 - be running Kubernetes version 1.14.x, 1.15.x, or 1.16.x
 - have a minimum of 5 nodes
 - have a minimum of 3 CPU, 7.5GB memory per node
-- have a metrics-server
-  - This may be as simple as running something like `kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml`
-
-### IaaS Requirements
```
jamespollard8, line 42: hmm, I actually liked this section and the formatting it had before

docs/deploy.md
```diff
@@ -140,24 +135,38 @@ This project is in it's early stages of development and hence there are features
 
       1. Update the `gcp_project_id` portion to your GCP Project Id.
       1. Change `contents_of_service_account_json` to be the entire contents of your GCP Service Account JSON.
-</br>
```
jamespollard8, line 143: I think deleting this causes major unwanted changes to the markdown

docs/deploy.md
```diff
@@ -140,24 +135,38 @@ This project is in it's early stages of development and hence there are features
...
+      > cf-for-k8s uses [ytt](https://github.com/k14s/ytt) to create and maintain reusable YAML templates. You can visit the ytt [playground](https://get-ytt.io/) to learn more about it's templating features. 
+      > In the above command, `ytt` can take a folder e.g. `config` or file via `-f`. See all options by running `ytt help`.
+    
+      ii. Install using `kapp` and pass the above K8s configuration file
+      ```console
+      kapp deploy -a cf -f /tmp/cf-for-k8s-rendered.yml -y
+      ```
+      > cf-for-k8s uses [kapp](https://github.com/k14s/kapp) to manage it's lifecycle. `kapp` will first show you a list of resources it plans to install on the cluster and then will attempt to install those resources. `kapp` will not exit untill all resources are installed and their status is running. See all options by running `kapp help`.
 
+   Once you run the command, it should take about ~10-15mins dependning on your cluster bandwidth, size. `kapp` will provide updates on pending resource creations in the cluster (pods, lbs) and will wait till all resources are created and running. Following shows a sample output from `kapp`.
```
jamespollard8, line 154: I'd change this to:

Once you run the command, it should take about 10 minutes depending on your cluster bandwidth, size. `kapp` will provide updates on pending resource creations in the cluster and will wait until all resources are created and running. Here is a sample snippet from `kapp` output:

docs/deploy.md
Syerram: @jamespollard8 
The cluster requirements are the same as IaaS requirements, hence I collapsed them into one. In fact, there is no such thing as IaaS, it's just does your cluster provider support these requirements. 

The metrics server will go away once the metrics team will package it with cf-for-k8s.

docs/deploy.md
Syerram: Are you sure? I did a view file and I saw the numbering to be accurate.

jamespollard8: Ah cool, you're right 👍

PR #179, opened 2020-05-04
Use tls by default for application ingress

### LINKED ISSUE #46: Apps not available via HTTPS

XanderStrike:
- This enables tls by default for the workload domain
- It updates generate values, but does not enforce different apps and system domains
  - When app and system domain are the same, system domain is used (this is what generate values does)
  - When app and system domain are different, separate certs are used

**TL;DR: This implementation no longer requires app and system domains to be different.**

fixes #46

**Acceptance Steps**

```
GIVEN a kubernetes cluster
WHEN I deploy cf-for-k8s
AND push an application
THEN I can reach that application via https
```

@jrussett @jenspinney @ndhanushkodi @tcdowney @kauana


config/_ytt_lib/github.com/cloudfoundry/cf-k8s-networking/config/cfroutesync/istio-gateway.yaml
```diff
@@ -1,17 +0,0 @@
-#@ load("@ytt:data", "data")
----
```
tcdowney, line 2: Why are we switching the apps over to using the `cf-system` Gateway?

I think we'll also need to update these docs: https://github.com/cloudfoundry/cf-k8s-networking/blob/master/doc/managing-certificates-for-application-ingress.md

XanderStrike: Having multiple gateways doesn't buy us anything. If two gateways have conflicting configuration Istio will silently accept them and then generate invalid configuration to ship to the ingress gateways. Better in my opinion to have a single gateway as the source of truth where we can see if we've accidentally tried to make two `servers` with the same wildcard host.

Also, app domains aren't necessarily tied to a given space or org, so in the ideal future where spaces are mapped to namespaces it would leave us with a complicated tangle of gateways.

config/external-routing.yml
```diff
@@ -24,32 +36,8 @@ metadata:
     kapp.k14s.io/change-group: istio.io/ingressgateway
 
 #! this gateway is used to expose CF services in the cluster, externally
```
jamespollard8, line 38: might be nice to include this in-line comment at the top of the gateway.lib.yml file too

config/gateway.lib.yml
```diff
@@ -0,0 +1,39 @@
...
+kind: Gateway
+metadata:
+  name: istio-ingressgateway
+  namespace: #@ system_namespace
+spec:
+  selector:
+    istio: ingressgateway
+  servers:
+  - hosts:
+    #! TODO: restrict this
```
jamespollard8, line 12: Are you planning to address this TODO as part of this PR? If not, do you have a story to track this?

jrussett: @jamespollard8, @XanderStrike and I wrote [this story](https://www.pivotaltracker.com/story/show/172662780) in our tracker to figure out how we wanted to deal with this `TODO`.  We did not plan on addressing the comment in this PR.

jamespollard8: Ah, on first pass I didn't catch that this TODO was moved, not created

jenspinney: Also, just in case it wasn't clear, the TODO wasn't added in this PR. This was in the old external-routing.yml that this was extracted from.

(I believe I added it back in the early days of cf-for-k8s when we were just racing to get anything working at all).

PR #180, opened 2020-05-04
Remove entries from sample values that are not valid anymore

Syerram:
- Removed `log_cache_client` after logging 2.0 was introduced
- Moved the metric proxy above registry


jamespollard8:
this was taken care of in 9b1aec7

PR #186, opened 2020-05-06
Update blobstore region default to be empty string

piyalibanerjee:
In response to suggestions made by Rel-Int team here: https://github.com/cloudfoundry/capi-k8s-release/issues/24, CAPI made the following changes:
 
-blobstore.region default value was previously "''". This configurable value has been changed to default to an empty string ("") instead.
- `capi_kpack_watcher` image property has been added to config/values.yml as well, defaulting to an empty string just like the other images. 

Authored-by: Piyali Banerjee <pbanerjee@pivotal.io>

>

---

**Acceptance Steps**


piyalibanerjee:
The PR branch was mistakenly based on master rather than on develop. I'm closing the PR and will re-open with a branch based on develop.

PR #191, opened 2020-05-13
ENH: add hack script for validating app_registry credentials

jamespollard8:
If these credentials are misformatted or invalid, you may not find out until after the kapp deploy finishes and you try to push an app.

The idea here is that you can run this script upfront to validate the credentials and tighten this feedback loop.

Known issues / future enhancements
- This commits a minimal Dockerfile to the repo, which likely is not necessary.
- This pushes a docker image to a repository and doesn't fully clean up after itself.
- This cannot yet handle Azure container registry (only docker and gcr)

```
$ hack/validate-registry-access.sh --help
...

Confirm that app_registry credentials have push access to the specified repository.

usage:
  app_registry_check.sh <yaml with app_registry configuration>
```

---

**Acceptance Steps**

Scenario: valid docker creds
When I run `app_registry_check.sh <yaml with app_registry configuration>`
And the creds in the file are valid
Then the script exits with exit code zero.

Scenario: valid GCR creds
When I run `app_registry_check.sh <yaml with app_registry configuration>`
And the creds in the file are valid
Then the script exits with exit code zero.

Scenario: bad docker creds
When I run `app_registry_check.sh <yaml with app_registry configuration>`
And the password in the file is wrong
Then the script exits with non-zero exit code.

Scenario: malformed gcr creds
When I run `app_registry_check.sh <yaml with app_registry configuration>`
And the password in the file is malformed json
Then the script exits with non-zero exit code.

### DIFF
--- a/hack/app-registry-check-dockerfile/Dockerfile
+++ b/hack/app-registry-check-dockerfile/Dockerfile
@@ -1,0 +1,2 @@
+FROM scratch
+CMD ["echo", "hello"]
--- a/hack/validate-registry-access.sh
+++ b/hack/validate-registry-access.sh
@@ -1,0 +1,81 @@
+#!/usr/bin/env bash
+echo "WARNING: The hack scripts are intended for development of cf-for-k8s.
+They are not officially supported product bits. Their interface and behavior" \
+"may change at any time without notice." 1>&2
+
+# This is a hack! see https://github.com/cloudfoundry/cf-for-k8s/blob/develop/hack/README.md
+set -eo pipefail
+
+# requirements
+# General: docker cli, yq
+# GCR: run `gcloud auth configure-docker`
+#      gcloud must be in the path
+
+show_help() {
+    echo ""
+    echo "Confirm that app_registry credentials have push access to the specified repository."
+    echo ""
+    echo "usage:"
+    echo "  app_registry_check.sh <yaml with app_registry configuration>"
+    echo ""
+    exit 1
+}
+
+docker_tag_and_push() {
+    echo "building tiny test docker image..."
+    docker build --tag ${docker_tag} $(dirname $0)/app-registry-check-dockerfile/ 1>/dev/null
+
+    echo "docker push-ing ${docker_tag} to test push access..."
+    docker push ${docker_tag} 1>/dev/null
+
+    docker rmi ${docker_tag} 1>/dev/null
+    echo "Confirmed push access to $1"
+}
+
+main() {
+
+    values_file="$1"
+
+    if [[ -z ${values_file} || ${values_file} = "-h" || ${values_file} = "--help" ]]; then
+        show_help
+    fi
+
+    if [[ "$(yq -r .app_registry $values_file)" == "null" ]]; then
+        echo "No app_registry key found at top level of ${values_file}"
+        exit 1
+    fi
+
+    registry_host="$(yq -r .app_registry.hostname ${values_file})"
+    username="$(yq -r .app_registry.username ${values_file})"
+    repo="$(yq -r .app_registry.repository ${values_file})"
+    docker_tag="${repo}/cfk8s-test-delete-me"
+
+    if [[ $registry_host = "https://index.docker.io/v1/" ]]; then
+        password="$(yq -r .app_registry.password ${values_file})"
+
+        echo "logging into dockerhub with username and password"
+        docker login -u "${username}" -p "${password}"
+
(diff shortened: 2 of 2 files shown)


Birdrock:
@jamespollard8 Some changes were made to the app domain, so you'll need to rebase your PR for retesting.

davewalter:
@jamespollard8 Given that we want to eliminate the `hack` directory entirely, I think it would be great if we could find a way to do this that didn't involve adding a hack script. What is the ultimate goal of this script? Am I correct in assuming that it will be used by customers and therefore part of the officially supported product?

drnic:
What do you mean by “customers” with respect to this open source project?

Scripts and documentation that help users fail fast are a good thing for the users and for the maintainers who might get “this doesn’t work” issues filling github and slack, then the error might have been presented to the user clearly upfront; rather than obscurely by “cf push” failing 10 minutes later.

It is desirable to encourage “fail fast” solutions for many reasons. If you don’t want helper/hack scripts, how will we help users to ensure they’ve constructed their yaml files correctly?

drnic:
> Am I correct in assuming that it will be used by customers and therefore part of the officially supported product?

Please don’t make this distinction.

jamespollard8:
Hey @drnic as I mentioned [on slack](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1589401941333300?thread_ts=1589400289.333100&cid=CH9LF6V1P, I think you’ve misunderstood Dave’s comments here. The customers and product that he’s referring to are all cf-for-k8s focused - nothing closed-source. The distinction he's drawing is more between beta cf-for-k8s and GA'd cf-for-k8s.

davewalter:
Sorry for the confusion @drnic. As @jamespollard8 mentioned, by "customers", I meant "users", and by "officially supported product", I was trying to emphasize something that would be supported by RelInt if a user ever had a problem with it, as opposed to something that we use internally and have decided to make public in the hopes that others find it useful, but that we reserver the right to change at any moment without prior notification.

I hope that helps clarify my original statement.

drnic:
Thanks for clarification, sorry about that.

Re: “getting rid of hacks directory” — is the plan to replace the `bosh int` wrapper script to generate secrets, and to replace other scripts (I don’t think I’ve used the other hack/ scripts) mentioned somewhere I can read? Will these go into ytt or similar? Or will secrets generation become internal to K8s via k14s secrets-controller?

On this PR’s topic of failing fast when in outs are bad, how can we help users without resorting to wrapper or helper scripts? Cleverer features in ytt or kapp?

ericpromislow:
[finishes #172822213](https://www.pivotaltracker.com/story/show/172822213)
Co-authored-by: awittrock@pivotal.io

jamespollard8:
Oh @ericpromislow @Birdrock , FWIW I was expecting more conversation here and was going to wait on more consensus before hopefully hitting the merge button myself.

Syerram:
@jamespollard8 Thanks for taking the initiative to build this script. We have noticed users often trip on this step.

I would recommend that we flush out the use cases and problems that we are trying to solve. For e.g. do we recommend users to run this once before installation? or every time when they upgrade or change registry settings? Will we provide a similar interface for a future external database, blobstore, etc. 

There are other situations where `kpack` may not be able to push to the registry. for e.g. users configured istio to prevent any off cluster access and hence kpack fails to push to registry.    

IMO, the script is solving a symptom and not the core problem. I don't see an issue that the user sees a cf-push fail due to registry settings. However, it is a pain to understand why cf push failed and even more painful to update the running foundation once you have the fix (Today, you have to delete and re-install cf-for-k8s). A better approach is
- `cf push` should print an actionable error message (e.g. "Unable to push image `app:sha` to registry due invalid user/password")
- The user can update the existing foundation by re-running the kapp deploy with the right registry settings. Assuming cf-for-k8s correctly, the user will get real feedback from the cluster in minutes.

I do see value in performing syntactic checks (e.g. is docker.io URL is correct) over semantic checks (is the URL actually reachable and usable). At a minimum, we should consider adding syntactic checks to YTT template (via functions I think).

jamespollard8:
Use case: the GCR creds I'm using expired. Currently, this manifests as a failure to push an app from source code. I'd like to run a quick check before running kapp deploy to validate that my creds actually have push access to the specified registry.

I like the idea of using more syntactic checks where possible. One problem I see is that it doesn't make sense conceptually to run "unit" tests on a values file.

This whole concept is something we'll continue to check back in on at various milestones, for example making the `app_registry` section of values a requirement.

PR #202, opened 2020-05-20
bump networking images

ndhanushkodi:
The new images include a label describing the repository and commit SHA that the image is built from

[#172922921](https://www.pivotaltracker.com/story/show/172922921)

---

**Acceptance Steps**
**When** I run `docker inspect <IMAGE>`, replacing `<IMAGE>` with the cfroutesync image that's deployed as part of cf-for-k8s.

**Then** I can see a label called "io.pivotal.metadata" with information describing the dependencies packaged into the image, including the github repository and commit SHA used to build the image

Note: The "io.pivotal.metadata" label is temporary and due to the tooling we're using at the moment. We plan to update this label to be more vendor agnostic as soon as we can get the tooling changes and a standardized label for cf-for-k8s.

@jenspinney @keshav-pivotal 

cc @evanfarrar


jamespollard8:
Sweet - looks good to me

PR #203, opened 2020-05-20
Annotate metric-proxy image

heycait:
[#172921741]

Signed-off-by: Louie Brann <lbrann@pivotal.io>

@louis-brann @hev

### DIFF
--- a/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
+++ b/config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
@@ -4,7 +4,7 @@ system_domain: ""
 system_namespace: ""
 
 images:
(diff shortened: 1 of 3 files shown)


acosta11:
Hello,

We docker inspected the provided image reference, but are having trouble finding the deplab labels. The output includes `            "Labels": null` in the Config section where we expected to see the large block of yaml describing the git sha used to build and all of the linux utils in the dockerfile. Would y'all happen to have annotated a different image than the one specified here?

Thanks,
Andrew and @davewalter

heycait:
@acosta11 I think we're a little confused. We ran ` deplab inspect --image oratos/metric-proxy:latest` after running the other `deplap` command and our image and see a bunch of what looks like OSL things.

heycait:
We did it again and think it worked this time.

acosta11:
The giant json blob is there! Looks good!

Thanks,
Andrew and @davewalter

Issue #210, opened 2020-05-26
cf push fails in staging process


msathe-tech:
### Summary

_In your own words, describe the issue._
Setup CF on TKG/AWS. CF push fails in staging process.

### Reproduction Steps
[cf-for-k8s sha hyperlink here]
cf push test-node-app -p tests/smoke/assets/test-node-app

Error
   Running 'npm install'
   node_modules      -> Not found
   npm-cache         -> Not found
   package-lock.json -> Not found
   Selected NPM build process: 'npm install'
   Executing build process
   Completed in 783ms
   Configuring environment
   **NPM_CONFIG_LOGLEVEL   -> error**
   NPM_CONFIG_PRODUCTION -> true
   PATH                  -> $PATH:/layers/org.cloudfoundry.npm/modules/node_modules/.bin
   Reusing layers from image 'index.docker.io/msathepivotal/9c2c4b22-2d2b-41f8-831e-a955f48889bd@sha256:a1e65a9cbca362f7b1fec2c7342bbe5d006ff5a7d4ecf16558e47ab7e87dcf0f'
   Reusing layer 'launcher'
   Reusing layer 'org.cloudfoundry.node-engine:node'
   Reusing layer 'org.cloudfoundry.npm:modules'
   Reusing 1/1 app layer(s)
   Reusing layer 'config'
   *** Images (sha256:8b59fc5890e5c28190eeb3ea89b95da7ef6a360cacbf0826d1dfca3035048247):
   msathepivotal/9c2c4b22-2d2b-41f8-831e-a955f48889bd
   index.docker.io/msathepivotal/9c2c4b22-2d2b-41f8-831e-a955f48889bd:b2.20200526.184838
   Adding cache layer 'org.cloudfoundry.node-engine:node'
   Adding cache layer 'org.cloudfoundry.npm:npm-cache'
   Build successful

Waiting for app to start...

**Stats unavailable: Stats server temporarily unavailable.
FAILED**

[Please include the `kapp deploy...` command, including all `config-optional` files]

[IaaS name(s) here (GKE, AKS, EKS, minikube, kind, etc)]

output of `ytt --version ; kapp --version ; kubectl version`:
[paste output here]

[Special configuration here]

_What steps/actions led to the issue?_

_It's helpful to include snippets of the error response or logs output_

### Expected behavior

_What was the expected result?_

acosta11:
Hello,

[...]

Issue #177, opened 2020-05-03
Metrics fail to display after rolling application update


zrob:
After pushing an app it is possible to view metrics using `cf app my-app`

<img width="542" alt="Screen Shot 2020-05-03 at 3 48 11 PM" src="https://user-images.githubusercontent.com/1703742/80927896-a6e54000-8d55-11ea-8010-1260da02020a.png">

A followup `cf push` to update the app also eventually displays metrics as shown above.

However, using cf7 to `cf push my-app --strategy rolling` the metrics will never display as anything other than zero.

<img width="519" alt="Screen Shot 2020-05-03 at 3 54 52 PM" src="https://user-images.githubusercontent.com/1703742/80928012-823d9800-8d56-11ea-8d61-5f090ff12cbf.png">

Note that when reproducing the successful case it is necessary to wait for metrics to display due to metric scraping intervals.

Syerram:
Metrics team is reviewing this issue => https://www.pivotaltracker.com/story/show/172660947

Tagging @hev

kkburr:
After investigating, we discovered that the metrics-server stores system metrics for pods by guid, not app_guid (which is what our code was selecting)

`cf push` -ing multiple times creates a new pod for the app instance, but it sends the same app_guid and guid values. metric-proxy is matching on app_guid and was able to find the pod.

`cf7 push` -ing multiple times does the same thing as above ^

`cf7 push` -ing multiple times with --strategy rolling creates a new pod with a new app_guid, but preserves the original guid. Switching our code to select on guid finds the pod metrics; using app_guid never finds it.

Our team will PR this into cf-for-k8s today.

jamespollard8:
Closed by #195

PR #213, opened 2020-06-01
Optionally include metrics-server

louis-brann:
This PR allows providing metrics-server on clusters that don't provide it.

`metric_proxy` requires `metrics-server` to function, but not all IaaSes provide the metrics-server, so this change allows operators to include metrics-server on IaaSes that don't already provide it: (e.g. EKS, kind, minikube) using a config-optional file.

---

Relevant stories/issues:
- https://www.pivotaltracker.com/story/show/172821078
- https://www.pivotaltracker.com/story/show/172771099  
- https://www.pivotaltracker.com/story/show/172319933

- https://github.com/cloudfoundry/cf-for-k8s/issues/210#issue-625110008
- https://github.com/cloudfoundry/cf-for-k8s/issues/177 
- https://github.com/cloudfoundry/cf-for-k8s/issues/123 

**Acceptance Steps**

In an IaaS without a metrics-server (e.g. EKS),  include `config-optional/add-metrics-server-components.yml` when rendering the k8s template with `ytt`.

Once this is successfully deployed, there should be a metrics-server in the kube-system namespace, and you should see container metrics when running `cf app`.

Note: If deploying on `kind` or `minikube`, follow the steps in [deploy-local.md](https://github.com/cloudfoundry/cf-for-k8s/blob/metrics-server/docs/deploy-local.md), and you should see the same results: metrics-server running in kube-system and container metrics when running `cf app`.

pair: @heycait @kkburr 
PM: @hev 
team: @cloudfoundry/cf-loggregator

### DIFF
--- a/config-optional/add-metrics-server-components.yml
+++ b/config-optional/add-metrics-server-components.yml
@@ -1,3 +1,4 @@
+#! https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
--- a/config-optional/patch-metrics-server.yml
+++ b/config-optional/patch-metrics-server.yml
@@ -1,0 +1,21 @@
+#@ load("@ytt:overlay", "overlay")
+
+#! Patch metrics-server to work in Kind
+#! https://github.com/kubernetes-sigs/kind/issues/398#issuecomment-621143252
+#! https://github.com/kubernetes-sigs/metrics-server/issues/131#issuecomment-618671827
+
+#@overlay/match by=overlay.subset({"kind":"Deployment","metadata":{"name":"metrics-server", "namespace":"kube-system"}})
+---
+spec:
+  template:
+    spec:
+      containers:
+        #@overlay/match by="name"
+        - name: metrics-server
+          #@overlay/replace
+          args:
+            - --cert-dir=/tmp
+            - --secure-port=4443
+            - --kubelet-insecure-tls
+            - --kubelet-preferred-address-types=InternalIP
+
--- a/docs/deploy-local.md
+++ b/docs/deploy-local.md
@@ -22,7 +22,6 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
 
 - have a minimum of 6 CPU, 6GB memory if using 1 node
   - commonly configured via Docker Desktop > Preferences > Resources
-- have a running metrics-server (this is an important consideration for **Kind** or **kubeadm** clusters. You can see one way to install it in the [Kind deploy instructions](#steps-to-deploy-on-kind))
 
 ## Considerations
 
@@ -48,19 +47,18 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
 
 1. Follow the instructions in [Deploying CF for K8s](deploy.md).
 
-   - Include the [remove-resource-requirements.yml](../config-optional/remove-resource-requirements.yml) and
-     [remove-ingressgateway-service.yml](../config-optional/remove-ingressgateway-service.yml)
+   - Include the [remove-resource-requirements.yml](../config-optional/remove-resource-requirements.yml),
+     [remove-ingressgateway-service.yml](../config-optional/remove-ingressgateway-service.yml),
+     [add-metrics-server-components-v0.3.6.yml](../config-optional/add-metrics-server-components-v0.3.6.yml) and
+     [patch-metrics-server.yml](../config-optional/patch-metrics-server.yml)
      overlay files in the set of templates to be deployed. This can be achieved by
      using the following commands:
 
      ```console
      TMP_DIR=<your-tmp-dir-path> ; mkdir -p ${TMP_DIR}
-     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f <cf_install_values_path> > ${TMP_DIR}/cf-for-k8s-rendered.yml
+     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f config-optional/add-metrics-server-components.yml -f config-optional/patch-metrics-server.yml -f <cf_install_values_path> > ${TMP_DIR}/cf-for-k8s-rendered.yml
      kapp deploy -a cf -f ${TMP_DIR}/cf-for-k8s-rendered.yml -y
      ```
-
-1. Make sure you've installed a metrics-server.
-   - this may be as simple as running something like `kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml`
 
 1. Once the `kapp deploy` succeeds, you should be able to run `cf api api.vcap.me --skip-ssl-validation`, etc
 
@@ -70,6 +68,12 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
 
    ```console
    minikube start --cpus=4 --memory=8g --kubernetes-version=1.16.8 --driver=docker
+   ```
+
+1. Enable metrics-server.
+
+   ```console
+   minikube addons enable metrics-server
    ```
 
 1. Obtain minikube IP.
--- a/docs/deploy-local.md
+++ b/docs/deploy-local.md
@@ -49,7 +49,7 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
 
    - Include the [remove-resource-requirements.yml](../config-optional/remove-resource-requirements.yml),
      [remove-ingressgateway-service.yml](../config-optional/remove-ingressgateway-service.yml),
-     [add-metrics-server-components-v0.3.6.yml](../config-optional/add-metrics-server-components-v0.3.6.yml) and
+     [add-metrics-server-components.yml](../config-optional/add-metrics-server-components.yml) and
      [patch-metrics-server.yml](../config-optional/patch-metrics-server.yml)
      overlay files in the set of templates to be deployed. This can be achieved by
      using the following commands:
--- a/tests/configs/configs_test.go
+++ b/tests/configs/configs_test.go
@@ -101,18 +101,9 @@ var _ = Describe("Configs", func() {
 
 			configDirectory := filepath.Join(filepath.Dir(filepath.Dir(currentDirectory)), "config-optional")
 
-			configs := []string{}
-			filepath.Walk(configDirectory, func(path string, info os.FileInfo, err error) error {
-
-				basename := info.Name()
-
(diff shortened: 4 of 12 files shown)


config-optional/patch-metrics-server.yml
```diff
@@ -0,0 +1,20 @@
...
+    spec:
+      containers:
+        #@overlay/match by="name"
+        - name: metrics-server
+          #@overlay/replace
+          args:
+            - --cert-dir=/tmp
+            - --secure-port=4443
+            - --kubelet-insecure-tls
+            - --kubelet-preferred-address-types=InternalIP
```
jamespollard8, line 20: nit: missing newline

tests/configs/configs_test.go
```diff
@@ -43,6 +43,13 @@ var _ = Describe("Configs", func() {
 		command.Wait()
 		print(" [done]\n")
 
+		print("enabling metrics server...")
+		command = exec.Command("sed", "-i", "s/enable_metrics_server: false/enable_metrics_server: true/g", "/tmp/dummy-domain-values.yml")
```
jamespollard8, line 47: Hmm, I don't love this but it makes sense. Should we maybe do this only for the 1 relevant/required config-optional file?

docs/deploy-local.md
```diff
@@ -49,19 +48,18 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
 
 1. Follow the instructions in [Deploying CF for K8s](deploy.md).
 
-   - Include the [remove-resource-requirements.yml](../config-optional/remove-resource-requirements.yml) and
-     [remove-ingressgateway-service.yml](../config-optional/remove-ingressgateway-service.yml)
+   - Enable the `metrics-server` by setting `metric_proxy.enable_metrics_server: true` in <cf_install_values_path>.
```
zrob, line 51: Any consideration to making an option in generate-values for this?  it could be more scriptable that way, rather than having to hand edit the file after the fact

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
```diff
@@ -7,6 +7,7 @@ images:
   metric_proxy: "oratos/metric-proxy@sha256:0e933b4f7208d3d6453ee28044126392a92055a565288434f2c522bf18a44fe1"
```
acosta11, line 7: > Fixing container metrics
Previously metric_proxy failed to get container metrics when using cf7 push --strategy rolling. Now metric_proxy uses the guid instead of the app_guid to get this correctly.

As a sanity check here, did we already happen to pull in the version of metric-proxy with this fix? I'm not seeing a diff on the image reference.

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/manifests/0-metrics-server-components-v0.3.6.yml
```diff
@@ -0,0 +1,157 @@
+#@ load("@ytt:data", "data")
+
+#@ if data.values.metric_proxy.enable_metrics_server:
```
acosta11, line 3: What was the motivation behind packaging all of these RBAC resources directly in metric-proxy instead of the config-optional file?

I worry whether the config-optional file functions without any of these permissions, but maybe I could see this organized as a separate file that works with metrics-server on any platform? So this file remains constant while the other config-optional file may change image or something based on the platform you have to install metrics-server it on?

tests/configs/configs_test.go
heycait: You mean just for the "check optional configs" test? Or a new section?

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
heycait: That could be the case. We assumed it hadn't made it in yet, can't remember if we made a PR to update it before but it was probably included when we annotated the image. We haven't switched over to cutting a release of metric proxy yet that will make it easier for us to track the image we're using.

docs/deploy-local.md
heycait: We hadn't thought about it but could take a look

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/manifests/0-metrics-server-components-v0.3.6.yml
heycait: Didn't think to package it in config-optional since that seemed to be a place for configs that mostly only apply to Kind environments. 

As was tested above here: https://github.com/cloudfoundry/cf-for-k8s/pull/213#discussion_r434881569, this shouldn't be applied if the metrics-server isn't added directly in the config dir. It'll fail the `ytt` compilation step.

docs/deploy-local.md
heycait: added `--enable-metrics-server` or `-m` flag

config/_ytt_lib/github.com/cloudfoundry/metric-proxy/config/values.yml
zrob: metrics have been appearing after cf push --strategy rolling for me for a couple weeks now

docs/deploy-local.md
jamespollard8: Heads up @zrob and @heycait, we're generally working towards reducing the amount of code/logic inside the generate-values script as we're aiming to one day get rid of it entirely.

hev:
@jamespollard8 metrics-server just works on most cluster providers - PKS, GKE, AKS, and EKS. The opt-in is for TKG (Alexandria edition), Kind, or Mini-Kube.

heycait:
Hey @jamespollard8 if we go the route of having a metrics-server that "just works", that would mean always including a metrics-server. Due to collisions with K8s API groups, that would mean we would now need our own K8s API group for the metrics-server. This would also mean there would be 2 metrics-servers in clusters that already provide them (i.e. GKE and AKS)—ours and the one by [kubernetes-sigs ones](https://github.com/kubernetes-sigs/metrics-server).

jamespollard8:
After a quick chat, we've decided that the next step is the logging/metrics team will try reimplementing this purely as config-optional files (not involving the cf-values file).

relint will make sure to check-in with our PM about the suggested interface before merging

config-optional/add-metrics-server-components.yml
```diff
@@ -0,0 +1,152 @@
+# https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml
```
ericpromislow, line 1: This needs to be `#! https…` for ytt

jamespollard8:
> it would be great to have a scriptable command that could deploy metrics server.
> 
> if a version has to be included in the command then any scripts will break when that update happens.

@zrob Are you suggesting a variant of `kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.6/components.yaml` just without a hard-coded version?

docs/deploy-local.md
```diff
@@ -48,19 +47,18 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
...
+     [add-metrics-server-components-v0.3.6.yml](../config-optional/add-metrics-server-components-v0.3.6.yml) and
+     [patch-metrics-server.yml](../config-optional/patch-metrics-server.yml)
      overlay files in the set of templates to be deployed. This can be achieved by
      using the following commands:
 
      ```console
-     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f <cf_install_values_path> > ./tmp/cf-for-k8s-rendered.yml
-     kapp deploy -a cf -f ./tmp/cf-for-k8s-rendered.yml -y
+     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f config-optional/add-metrics-server-components-v0.3.6.yml -f config-optional/patch-metrics-server.yml -f <cf_install_values_path> > /tmp/cf-for-k8s-rendered.yml
+     kapp deploy -a cf -f /tmp/cf-for-k8s-rendered.yml -y
```
jamespollard8, line 59: please toggle this back to ./tmp

docs/deploy-local.md
```diff
@@ -48,19 +47,18 @@ In addition to the Kubernetes version requirement in [Deploying CF for K8s](depl
...
+     [remove-ingressgateway-service.yml](../config-optional/remove-ingressgateway-service.yml),
+     [add-metrics-server-components-v0.3.6.yml](../config-optional/add-metrics-server-components-v0.3.6.yml) and
+     [patch-metrics-server.yml](../config-optional/patch-metrics-server.yml)
      overlay files in the set of templates to be deployed. This can be achieved by
      using the following commands:
 
      ```console
-     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f <cf_install_values_path> > ./tmp/cf-for-k8s-rendered.yml
-     kapp deploy -a cf -f ./tmp/cf-for-k8s-rendered.yml -y
+     ytt -f config -f config-optional/remove-resource-requirements.yml -f config-optional/remove-ingressgateway-service.yml -f config-optional/add-metrics-server-components-v0.3.6.yml -f config-optional/patch-metrics-server.yml -f <cf_install_values_path> > /tmp/cf-for-k8s-rendered.yml
```
jamespollard8, line 58: and `./tmp` here too please

zrob:
> > it would be great to have a scriptable command that could deploy metrics server.
> > if a version has to be included in the command then any scripts will break when that update happens.
> 
> @zrob Are you suggesting a variant of `kubectl apply -f https://github.com/kubernetes-
sigs/metrics-server/releases/download/v0.3.6/components.yaml` just without a hard-coded version?

@jamespollard8 I think roughly yes

If I follow the instructions as provided in this pr i could write a script that has something roughly like this (excluding other optional files)

`ytt -f config -f config-optional/add-metrics-server-components-v0.3.6.yml -f config-optional/patch-metrics-server.yml`

if that version changes then that script goes kaput

I'd rather be able to say something like

`ytt -f config -f config-optional/add-metrics-server.yml -f config-optional/patch-metrics-server.yml`

Or even better

`ytt -f config -f config-optional/add-metrics-server`

To relate back to your example, it's actually what I have scripted today, I use latest to avoid the version bit

`kubectl apply -f https://github.com/kubernetes-sigs/metrics-server/releases/latest/download/components.yaml`

As a consumer of this repo I shouldn't have to think about the metric server version, b/c the one included should work fine with the cf i'm installing.

jamespollard8:
@zrob Oh yeah totally. I already gave review feedback about dropping the version from the config-optional filename, I think they'd just forgotten to update the docs changes accordingly 👍

kkburr:
@jamespollard8 , updated based on your feedback from Thursday.

jamespollard8:
Looks great! Thanks for all your good work and many round of review on this.

PR #214, opened 2020-06-04
Replace cfroutesync with routecontroller

ndhanushkodi:
Other highlights:
- bump cf-k8s-networking to release v0.0.6
- use kbld to reference Istio images by digest instead of tag
- use OSL-ed images for Istio components

Routecontroller is a Kubebuilder-based controller that takes
the place of cfroutesync. It operates on Route custom resources
that Cloud Controller makes and translates them into Istio
VirtualServices and Kubernetes Services. For more information see:
[this adr](https://github.com/cloudfoundry/cf-k8s-networking/blob/master/doc/architecture-decisions/0010-route-crd-and-kubebuilder-instead-of-metacontroller.md) and [this doc](https://docs.google.com/document/d/1DF7eTBut1I74w_sVaQ4eeF74iQes1nG3iUv7iJ7E35U/edit)

[#172921777](https://www.pivotaltracker.com/story/show/172921777)
---

We expect that anyone upgrading to this version would need to delete and recreate all of their routes, since the controller managing routes has been changed. It might be easier to just deploy a new cf-for-k8s starting from this version.

We plan to make a future PR to remove metacontroller from cf-for-k8s. [173165427](https://www.pivotaltracker.com/story/show/173165427)

**Acceptance Steps**
**Given** I have deployed cf-for-k8s
**When** I push an app with a route
**Then** I can curl the app and get a successful response. 
**And** `kubectl get pods -ncf-system` no longer shows the `cfroutesync` pod
 
@tcdowney  @keshav-pivotal 
 cc @cwlbraa


tcdowney:
> We expect that anyone upgrading to this version would need to delete and recreate all of their routes, since the controller managing routes has been changed.

This is broadly true. I wanted to clarify it a bit since y'all have started running upgrade tests (per this [Slack discussion](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1591315109009000)). The `routecontroller` requires a `Route` custom resource to exist for each route and it was only recently that Cloud Controller started creating/updating/deleting these. I believe cf-for-k8s release [`v0.0.3`](https://github.com/cloudfoundry/cf-for-k8s/releases/tag/v0.3.0) shipped with this version of Cloud Controller so upgrades from that might be fine...

But upgrades from older cf-for-k8s installations that had routes created _before_ Cloud Controller supported the `Route` CRD will need those old routes recreated via the `cf` CLI. Just wanted to call this out cause I think there's a good chance that the upgrade tests *won't* fail and that it's probably fine.

Birdrock:
@ndhanushkodi @tcdowney Deployed and tested; smoke tests passed and cfroutesync isn't present. Looks good. Thanks for the PR with well written acceptance criteria and 'release notes'.

```
kubectl get pods -A | grep route
cf-system        routecontroller-6cc744d66c-lt66f                                 2/2     Running     4          10m
```

PR #218, opened 2020-06-04
Rename system certificate secret to cf-system-cert

christianang:
This PR renames the system certificate from `istio-ingressgateway-certs` to `cf-system-cert`. This is to support hot reloading of certificates when the secret changes. For some reason Istio doesn't hot reload certificates if the Kubernetes secret name begins with `istio`. The only reference I could find to this in their documentation is in a note [here](https://istio.io/docs/tasks/traffic-management/ingress/secure-ingress/#configure-a-tls-ingress-gateway-for-a-single-host), but it doesn't really explain why. It seems best to change it to a new name that doesn't start with `istio` and as a bonus this will also keep it consistent with the `cf-workloads-cert` secret name.

Related tracker story [#173181160](https://www.pivotaltracker.com/story/show/173181160).

---

**Acceptance Steps**

Given you have a cf-for-k8s environment deployed with the changes
You can

1. Ask the Istio Ingress Gateway for the certificate for a system component.
```
echo | openssl s_client -showcerts -servername api.<DOMAIN> -connect api.<DOMAIN>:443 2> /dev/null
```
2. Rotate your system certificate in `cf-values.yaml`.
  a. Assuming you created your environment with the `./hack/generate-values.sh`, you can open the `/tmp/<DOMAIN>/cf-vars.yaml` and delete the `system_certificate` key and values. Then rerun `./hack/generate-values.sh` again. At this point your `cf-values.yaml` should have a new certificate for the system certificate.

3. Deploy with the updated `cf-values.yaml`.

4. Ask the Istio Ingress Gateway for the certificate again and see the certificate matches the newly generated certificate under `system_certificate.cert` in `/tmp/<DOMAIN>/cf-vars.yaml`.
```
echo | openssl s_client -showcerts -servername api.<DOMAIN> -connect api.<DOMAIN>:443 2> /dev/null
```

@cloudfoundry/cf-networking

### DIFF
--- a/config/external-routing.yml
+++ b/config/external-routing.yml
@@ -6,7 +6,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: istio-ingressgateway-certs
+  name: cf-system-cert
   namespace: istio-system
   annotations:
     kapp.k14s.io/change-rule.istio-ingressgateway: "upsert before upserting istio.io/ingressgateway"
--- a/config/gateway.lib.yml
+++ b/config/gateway.lib.yml
@@ -23,7 +23,7 @@ spec:
       protocol: HTTPS
     tls:
       mode: SIMPLE
-      credentialName: istio-ingressgateway-certs
+      credentialName: cf-system-cert
   #@ non_system_domain_app_domains = [d for d in app_domains if d != system_domain]
   #@ if/end len(non_system_domain_app_domains) > 0:
   - hosts:


mike1808:
Maybe this will help

https://github.com/istio/istio/blob/41b5573553b6b8108febcb0779a42fd055f4b4e9/security/pkg/nodeagent/secretfetcher/secretfetcher.go#L155

>>>// isIngressGatewaySecret checks secret and decides whether this is a secret generated for ingressJimmy Chen, 1 year ago: • Skip validating non ingress gateway secret at…
// gateway. For secrets with prefix "istio" and "prometheus", they are generated by istio system and
// they are not gateway secrets. Other secrets generated by istio could have "token" field.
// isIngressGatewaySecret returns false if a secret has name prefix "istio", or "prometheus", or has
// "token" field.

Birdrock:
@christianang Is this the `cf-system-cert` or a `cf-ingress-cert`? Some of our terms are pretty overloaded, so I'm trying to make sure they convey the meaning we want them to.

@XanderStrike Does this require a cf-networking update? There are references to `istio-ingressgateway-certs` in github.com/cloudfoundry/cf-k8s-networking/config/istio-generated/xxx-generated-istio.yaml.

christianang:
It is the certificate used for the system domain by the ingress gateway e.g when I ask for `api.system-domain.com` this is the certificate returned.

This also won't require a cf-networking update. There are references to the `istio-ingressgateway-certs` in the Istio install by default, but from what I could tell there isn't actually a requirement to generate a certificate with the name `istio-ingressgateway-certs`.

Birdrock:
@christianang Thanks for the explanation. Sounds good to me. Thank you for your PR!

PR #219, opened 2020-06-08
Adding new istio config folders

ndhanushkodi:
- currently this will be a no-op to cf-for-k8s
- once we cut a new release for cf-k8s-networking, these folders will
have contents in them
- this is to keep cf-k8s-networking pipelines from breaking in the
meantime before cutting a release

- for more context see [this
comment](https://github.com/cloudfoundry/cf-k8s-networking/pull/47#issuecomment-640359538)

[#172031795](https://www.pivotaltracker.com/story/show/172031795)

---

**Acceptance Steps**

**Given** I have deployed cf-for-k8s before this commit
**When** I deploy cf-for-k8s with this commit
**Then** I can see that this is a no-op (nothing in the deployment should change)

@kauana @keshav-pivotal


jamespollard8:
Makes sense 👍

PR #221, opened 2020-06-09
Change docs to use ./tmp instead of /tmp

paulczar:
/tmp is ephemeral storage on most *nix system and a security risk on all

This updates the most common deploy instructions to use `./tmp` instead which
is excluded via `.gitignore`.

There's still an alarming amount of `/tmp` in use but this is a start.

>

---

**Acceptance Steps**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,5 @@
 **/*.tfstate.backup
 **/.terraform/
 **/*.tfvars
-tmp
+
 .idea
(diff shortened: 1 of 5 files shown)


jamespollard8:
We discussed some changes as a team and decided we didn't want to commit a tmp directory to cf-for-k8s but we also don't want to suggest using `/tmp` everywhere. The changes we settled on live in this commit: https://github.com/cloudfoundry/cf-for-k8s/commit/c5d8f3c8ce9262d952b6312124ec86cebdf23f77

Thanks again Paul for bringing this up and for the PR 👍

PR #223, opened 2020-06-10
Adds missing PSPs and associated RBAC controls for PSP enforced clusters

paulczar:
This PR enables deploying cf-for-k8s to a PSP enforced cluster in AWS.

As well as added the necesary PSPs/RBAC it also adds an optional ingress
configuration value for using Amazon NLB with a pre-defined IP, similar to
what already existing for istio static ips.

This is fully tested on a TKG cluster deployed to Amazon via Tanzu Mission Control.

>

---

**Acceptance Steps**


jpalermo:
Hey @paulczar. Thanks for the PR. We tested it and everything looks great.

However, after talking things over with our PM, it sounds like PSP is not a configuration we're ready to support and test at this point. Merging it in would mean either:
- We need to create the proper testing pipelines for it and support it as a first class feature
- We don't test it, and it likely breaks in the future and we don't know about it.

We'll be looking into a supported PSP configuration in the near future, but we don't currently have the bandwidth to support the first option, and we don't want potentially broken config either.

We're also going to work with the component teams to get some of this PSP config pushed down to them (where possible). Would the eirini PSP overlay make sense as a PR to them as a first step? It seems like they already have a PSP config.

jamespollard8:
> Would the eirini PSP overlay make sense as a PR to them as a first step?

Hey @paulczar, do you agree that most of the changes here could/should come in as upstream changes?

jamespollard8:
As a side note, we've decided that we want to move away from the config-optional model in favor of values files toggles that trigger conditional ytt templating. (for example: https://github.com/cloudfoundry/cf-for-k8s/pull/245/files#diff-13ace597115bdbee9dffbecf4839210bR5)

paulczar:
Yeah makes sense to both :)

davewalter:
Given the discussion above, I am going to close this PR without merging it. We look forward to receiving these changes via future component version updates.

Thanks again,
Dave and @jamespollard8

PR #244, opened 2020-06-17
Explicitly copy/paste default scopes for user

aashah:
* Allows for downstream overlays to safely append to the list of default scopes granted to a newly created UAA user, instead of being holding downstream overlays responsible for owning the default list.

---

**Acceptance Steps**

1. Deploy old cf-for-k8s
1. Using uaac, collect the list of scopes granted to a newly created user (`uaac user get <>`).
1. Deploy new cf-for-k8s
1. Using uaac, observe the list of scopes granted to a newly created user has not changed.

### DIFF
--- a/config/uaa.yml
+++ b/config/uaa.yml
@@ -132,6 +132,24 @@ oauth:
       authorities: cloud_controller.write,cloud_controller.update_build_state
       authorized-grant-types: client_credentials
       secret: #@ data.values.uaa.admin_client_secret
+  #@overlay/match missing_ok=True
+  user:
+    authorities:
+      - openid
+      - scim.me
+      - cloud_controller.read
+      - cloud_controller.write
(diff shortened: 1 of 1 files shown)


jamespollard8:
FYI @aashah because you have [or at least should have] push access, it's more convenient for us if you create PRs from branches because we do a bit more automated testing (ie install and run smoketests on GKE) on branches.

(we run fewer tests on forks for security reasons)

aashah:
Ahh, I totally knew that. I'm gonna just close both of my PRs and re-submit them under cf-for-k8s.

jamespollard8:
> Ahh, I totally knew that. I'm gonna just close both of my PRs and re-submit them under cf-for-k8s.

Cool cool - thanks. (FYI, we just changed the behavior so we now treat and test forks the same as branches)

PR #254, opened 2020-06-23
Enable tls for external capi database

c0d1ngm0nk3y:
Set the ca_cert value introduced in the capi release.

PR #258, opened 2020-06-26
Enable tls for external capi databases

kramerul:
This is a follow up of https://github.com/cloudfoundry/capi-k8s-release/pull/43 and https://github.com/cloudfoundry/cf-for-k8s/pull/253 to enable TLS for external databases.


kramerul:
This PR was already opened: https://github.com/cloudfoundry/cf-for-k8s/pull/254

Issue #262, opened 2020-06-30
Default-Deny Network Traffic for System Components


tcdowney:
**Is your feature request related to a problem? Please describe.**
We (cf-k8s-networking team) have a story ([#173371423](https://www.pivotaltracker.com/story/show/173371423)) around adding Layer 3 (L3) [`NetworkPolicy`](https://kubernetes.io/docs/concepts/services-networking/network-policies/) to default-deny traffic to/from system components. Components will need to explicitly allow traffic.

This will provide the following benefits:
* Improve the security posture of the platform. Additive allow policies will require conscious effort to enable components to interact with each other.
* These policies can help serve as living documentation around what network interactions are taking place between components. Similar to [these existing docs](https://github.com/pivotal-cf/docs-pcf-security/tree/2.9/networking), they'll show what protocols and ports are in use and where we expect traffic to be coming from.

**Describe the solution you'd like**
We will be adding `NetworkPolicy` _somewhere_ in `cf-for-k8s`. Right now it's kind of scattered all over the place (some lives in [eirini.yml](https://github.com/cloudfoundry/cf-for-k8s/blob/9e8556471b4b34910405423e9f69f941174960ee/config/eirini.yml#L21-L44) for example).

We propose **a central `config/network-policy.yml` file** that contains this configuration for all of the system components in `cf-for-k8`s so that an operator can see what is going to be applied at a glance. 

Alternatively components could manage these `NetworkPolicy` entries themselves and `cf-for-k8s` would just contain the initial "default-deny" policy. However, that feels more difficult to manage overall and poses release coordination issues.

These policies require coordination between various components (`podSelector`s and such need to line up) so having them managed in a central place feels valuable.

**Describe alternatives you've considered**
A good question was raised by @kauana around why not use Layer 7 (or L3 **and** L7) policies (in our case this would be [Istio `AuthorizationPolicy`](https://istio.io/latest/docs/reference/config/security/authorization-policy/#AuthorizationPolicy)) for this? We [discussed this on Slack a bit here](https://cloudfoundry.slack.com/archives/CFX13JK7B/p1593538184408400) and settled on L3 `NetworkPolicy` should be the minimum required amount of security provided. L7 policies can be layered on to provide additional security, but may not be required (or desired) for all components.

Some thoughts we had:

* L3 `NetworkPolicies` provide a good service mesh agnostic baseline level of protection<sup>1</sup>.
[...]

Issue #261, opened 2020-06-29
New Label needed on cf-system


kkburr:
**Is your feature request related to a problem? Please describe.**
When we have installed prometheus onto our cluster using helm, it currently cannot scrape any metrics from "cf pushed" app pods in the cf-workloads namespace because of the network policies in place that restrict ingress:

https://github.com/cloudfoundry/cf-for-k8s/blob/d9cf7e018d84bcf07088c94e86a1bf7cada6b253/config/eirini.yml#L26
https://github.com/cloudfoundry/cf-for-k8s/blob/d9cf7e018d84bcf07088c94e86a1bf7cada6b253/config/_ytt_lib/eirini/rendered.yml#L5

**Describe the solution you'd like**
After talking to @tcdowney, we agreed that the best solution would be to create a network policy that allows ingress from prometheus to a pod in cf-workloads. That requires 3 things:

- the new network policy (we will provide an example in our [examples repo](https://github.com/cloudfoundry/cf-for-k8s-metric-examples/tree/master)...currently on a branch to go into master today)
- prometheus to be installed with the new pod label expected in the network policy (and installed in the cf-system namespace since it has istio enabled already)
- the cf-system namespace to have the new namespace label expected in the network policy.

We need you all to take care of bullet point 3 by adding this label to the `cf-system` namespace:
`cf-for-k8s.cloudfoundry.org/cf-system-ns: ""`

Please see the thread with additional context: https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1593188901246400

**Describe alternatives you've considered**
We considered using existing labels defined in the network policies above ^ that would allow our traffic in, but this was a hack solution.

CC @heycait

kkburr:
We decided to change our network policy to match on `name: cf-system`:
```
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-prometheus-scrape
  namespace: cf-workloads
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          name: cf-system
      podSelector:
        matchLabels:
          what-am-i: prometheus
```

jamespollard8:
Fixed by fa7c634

Issue #168, opened 2020-04-29
Feature request: How can we support configuring an external database?


ciriarte:
### Summary

Deploying a database with `cf-for-k8s` is really neat, but I am wondering how can we support an external database that we can manage outside of the deployment lifecycle of `cf-for-k8s`.

### Expected behavior

My expectation is to set data.values with the appropriate config and have my installation point to my external database.

zankich:
This PR https://github.com/cloudfoundry/cf-for-k8s/pull/143 describes how to configure CC to use an external database.

loewenstein:
I've had a closer look at external databases for cf-for-k8s. It appears to me that a couple of things have to happen.
**First**, we need to include documentation in #143 on how to properly prepare the database, like it is done [for the internal database](https://github.com/cloudfoundry/cf-for-k8s/blob/d9bd7e07b5e0eb58fcc0fc227d11c556c2ab5d3b/config/postgres.yml#L65-L79).
If I have it correct, connections to external databases would currently be plain text though.
Hence, **second** we should include a warning about that in #143.
Having checked that for both capi and uaa, we should **third** add documentation for configuring the uaa to use an external database and **fourth** not include an internal database if neither capi nor uaa would use it.

When uaa gets bumped to https://github.com/cloudfoundry/uaa/commit/6f3c7fe3f726bf8d961c420ac7325697fdd71c92 with [#172805177](https://www.pivotaltracker.com/n/projects/1382120/stories/172805177) and capi gets support for configuring TLS with [#172661684](https://www.pivotaltracker.com/story/show/172661684), probably touching [capi-k8s-release/templates/ccng-config.lib.yml#L121](https://github.com/cloudfoundry/cf-for-k8s/blob/ce7a364c776cc82f6724b2d382398078d5fb0a07/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/ccng-config.lib.yml#L121) and [capi-k8s-release/templates/ccng-config.lib.yml#L124-L125](https://github.com/cloudfoundry/cf-for-k8s/blob/ce7a364c776cc82f6724b2d382398078d5fb0a07/config/_ytt_lib/github.com/cloudfoundry/capi-k8s-release/templates/ccng-config.lib.yml#L124-L125), we can **fifth** integrate tests for external databases into the Concourse pipeline and **sixth** remove any limitations for the documentation provided with #143.

[...]

PR #270, opened 2020-07-02
Initial NetworkPolicies for System Components

tcdowney:
This is a followup to the ["Default-Deny Network Traffic for System Components"](https://github.com/cloudfoundry/cf-for-k8s/issues/262) we logged the other day.

This PR includes some `NetworkPolicies` that lock deny all egress/ingress for the system components by default and have specific allow policies for known ingress/egress networking. The following diagram shows the interactions we mapped out and discovered through trial and error. It is **not exhaustive** however. We will need help from other component teams to fill in the gaps!

**Update:** The following diagram is now out of date:
system-component-network-interactions

## What We Need
We, the cf-k8s-networking team, don't and can't know all the possible interactions between components. We need the teams that own these components to read over these policies and correct any mistakes that we've made on this initial attempt.

We also need teams to supply information on what **ports** and **protocols** their components use. As written these policies allow any port/protocol and we want to be explicit here.

**It would be awesome if you could just add additional commits to this branch to add these on your policies!**
https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource
 
## Issues
* The system component namespaces (`cf-system`, `cf-db`, `cf-blobstore`, `cf-workloads-staging`, etc.)  were lacking labels for us to reference. The same issue as https://github.com/cloudfoundry/cf-for-k8s/issues/261. We added some placeholder `name: <namespace-name>` labels in the meantime.
* All pods needed to be able to communicate with platform dependencies running in `kube-system` (such as kube-dns) so we added blanket allow rules for ingress/egress traffic related to that namespace. However, we had to label the namespace ourselves with `name: kube-system` and I'm not sure how something like this would work out of the box. Looking for input here.
* Components that needed to talk to the Kubernetes API (e.g. Eirini, metric-proxy, cf-api-kpack-watcher) or external services (e.g. Cloud Controller, UAA) were given wide-open egress policies. The [external services (identity providers, db, service brokers, etc.) concern](https://github.com/cloudfoundry/cf-for-k8s/issues/262#issuecomment-652056293) deserves its own story/focused effort. Same with the ability to talk to the Kubernetes API -- particularly "off-platform" control planes like on GKE (this is easy if your K8s API lives in `kube-system`).

## Other Thoughts
Writing this YAML by hand was pretty toil-heavy. @XanderStrike had an interesting idea where one might use `ytt` to encode structures representing each arrow on the graph. Like this:

```
#@data/values

---
connections:
- name: migrate-to-db
  fromNs: cf-system
  fromSelector:
    job-name: ccdb-migrate
  toNs: cf-db
  toSelector:
    app: postgresql
- name: api-worker-to-db
  fromNs: cf-system
  fromSelector:
    app.kubernetes.io/name: cf-api-worker
  toNs: cf-db
  toSelector:
    app: postgresql
```

And use ` ytt` loop to generate the `NetworkPolicy` based off of that.
```
#@ load("@ytt:data", "data")

#@ for conn in data.values.connections:
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: #@ "{}-egress".format(conn.name)
  namespace: #@ conn.fromNs
spec:
  podSelector:
    matchLabels: #@ conn.fromSelector
  policyTypes:
  - Egress
  egress:
  - to:
    - namespaceSelector:
        matchLabels:
          name:  #@ conn.toNs
      podSelector:
        matchLabels: #@ conn.toSelector
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: #@ "{}-ingress".format(conn.name)
  namespace: #@ conn.toNs
spec:
  podSelector:
    matchLabels: #@ conn.toSelector
  policyTypes:
  - Ingress
  ingress:
  - from:
    - namespaceSelector:
         matchLabels:
           name: #@ conn.fromNs
      podSelector:
        matchLabels: #@ conn.fromSelector
```

For the sake of this initial PR I wanted to keep the `NetworkPolicy` as close to how they're actually represented as possible, but would welcome feedback on this alternative approach. We want to make sure we end up with something that fits with the existing `cf-for-k8s` patterns here and is maintainable.

---

Thank you!
Tim && @cloudfoundry/cf-networking 

Tagging some component teams:
cc: @cloudfoundry/cf-capi @cloudfoundry/eirini @cloudfoundry/cf-loggregator @cloudfoundry/cf-metric-store


config/network-policies.yaml
```diff
@@ -0,0 +1,566 @@
...
+  name: fluentd
+  namespace: #@ data.values.system_namespace
+spec:
+  policyTypes:
+  - Egress
+  podSelector:
+    matchLabels:
+      app: fluentd
+  egress:
+  - {}
```
tcdowney, line 566: This is wide-open. Originally I had `fluentd` only allowed to egress to `log-cache`, but it was trying to make an `http` call to something. Would appreciate help on this one.

config/network-policies.yaml
```diff
@@ -0,0 +1,566 @@
...
+      namespaceSelector:
+        matchLabels:
+          name: #@ data.values.system_namespace
+    - podSelector:
+        matchLabels:
+          app: log-cache
+      namespaceSelector:
+        matchLabels:
+          name: #@ data.values.system_namespace
+  egress:
```
loewenstein, line 293: That's deny all, isn't it? This will collide with the [support for external databases](https://github.com/cloudfoundry/cf-for-k8s/issues/168). As the IPs of external databases might not be known or even fixed, I suppose we will have to allow egress traffic to all public IPs, potentially also Cloud Provider network internal IPs.

I would appreciate, if this doesn't get merged without a documented concept of how to open up the egress rules.
At the very least the [external database documentation](https://github.com/cloudfoundry/cf-for-k8s/blob/master/docs/configuration/external-databases.md) should be supported and updated.

loewenstein: Ah ok, RTFM. 
```yaml
egress:
  - {}
```

is actually allow all. As this is work in progress, the general comment about keeping external databases in mind still stands.

tcdowney: Yep, was keeping external databases, external service brokers, external identity providers, etc. in mind for Cloud Controller/UAA.

oppegard:
Hi @tcdowney, here's our (Log Store team) thoughts about this. Note that we're not a core component of CF (e.g. not in cf-deployment or available as an ops-file). We're coming from the viewpoint of a service that is tightly coupled to CF, that a customer may optionally deploy alongside cf-for-k8s. It is deployed in its own namespace, currently `log-store`.

1. We communicate with UAA and cf-api-server to determine if a user is authorized to view certain logs. We currently do this with the internal service names (uaa.cf-system.svc.cluster.local and capi.cf-system.svc.cluster.local). We added `istio-injection: enabled` to our namespace so our pods get the sidecar for handling mTLS to UAA, cf-api-server, and ourselves.

1. We use the cf-system/istio-ingressgateway for external ingress to our service (e.g. log-store.system.my-foundation.com).

1. Presently, we manually configure fluentd to egress logs to a service in our namespace. @pianohacker may have more info on how "syslog sinks" will work in cf-for-k8s.

Things are working well for us currently, but we're not sure what impact NetworkPolicies will have on the above points.

tcdowney:
@oppegard for points 1 and 2 you would need to include `NetworkPolicy` similar to this one for `log-cache`:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/files#diff-652a3e228db9cbb604d4ba5801be220fR484-R529

You could probably remove some of the `ingress` entries (like I don't think `cf-api-server` talks to your component), but that should serve as a decent example.

For the `fluentd` portion.. I think you're just configuring the `fluentd` that is deployed as a `DaemonSet` already right? The policies in this PR right now aren't restricting what `fluentd` can egress to. See:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/files#r449295249

docs/component_authors/networking.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Namespaces and Istio Sidecar Injection
+CF for Kubernetes uses Istio's [automatic mTLS](https://istio.io/latest/docs/tasks/security/authentication/authn-policy/#auto-mutual-tls) with a `STRICT` authentication policy. This means that all `Pods` that have an Istio sidecar will require any `Pods` that communicate with them to also have an Istio sidecar and use mutual TLS to secure traffic. Traffic that doesn't use mTLS or traffic that does not come from an Istio sidecar will be rejected and the connection will fail.
+
+When you create a new namespace, in order to take advantage of all of Istio's features, and communicate with other `Pods` that Istio sidecars, your namespace must have an `istio-injection=enabled` label to turn on automatic [Sidecar Injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/). This makes it so the pods in your namespace will contain an [Istio sidecar proxy](https://istio.io/latest/docs/reference/config/networking/sidecar/) running alongside them. In CF for Kubernetes, Istio sidecar injection is added automatically by a `ytt` [config](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/networking.yml#L63).
```
jamespollard8, line 118: nit: I'd say "..is added automatically by a `ytt` [overlay]

config/network-policies.yaml
```diff
@@ -0,0 +1,563 @@
...
+          cloudfoundry.org/source_type: STG
+      namespaceSelector:
+        matchLabels:
+          name: #@ data.values.staging_namespace
+    - namespaceSelector:
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
+      podSelector:
+        matchLabels:
+          app.kubernetes.io/name: cf-api-server
```
piyalibanerjee, line 410: We might need to ad the cf-api-worker to this since it may be used to delete from the blobstore.

config/network-policies.yaml
```diff
@@ -0,0 +1,563 @@
...
+metadata:
+  name: capi-kpack-watcher
+  namespace: #@ data.values.system_namespace
+spec:
+  policyTypes:
+  - Egress
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: cf-api-kpack-watcher
+  egress:
```
cwlbraa, line 464: right now, to my knowledge, the cf-api-kpack-watcher needs to talk to:

- k8s API
- cf-api-server
- container registry

config/network-policies.yaml
```diff
@@ -0,0 +1,563 @@
...
+      namespaceSelector:
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
+  egress:
+  - {}
+---
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: capi-kpack-watcher
```
piyalibanerjee, line 456: Heads up: this name will change soon after implementing this ADR: https://github.com/cloudfoundry/capi-k8s-release/commit/6863e894c580e44d5652a358fdd33867805de850

piyalibanerjee:
Hello! @cwlbraa and I looked over all the CAPI-related Network Policies. They seem right to us; we left a few comments on some of them.

config/network-policies.yaml
tcdowney: This config lets it talk to anything.

Anything that has to talk to the k8s API needs open egress to be generic because the Kubernetes API may be external to the cluster _and_ we don't have a generic way of knowing what IP address(es) it might be at.

That's not to mention the container registry bit.

config/network-policies.yaml
tcdowney: The name of the policy doesn't matter, it just needs to make sense to whoever is looking at it. If would like it to be named something differently feel free to update it! 👍

docs/component_authors/networking.md
```diff
@@ -0,0 +1,122 @@
...
+
+When you create a new namespace, in order to take advantage of all of Istio's features, and communicate with other `Pods` that Istio sidecars, your namespace must have an `istio-injection=enabled` label to turn on automatic [Sidecar Injection](https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/). This makes it so the pods in your namespace will contain an [Istio sidecar proxy](https://istio.io/latest/docs/reference/config/networking/sidecar/) running alongside them. In CF for Kubernetes, Istio sidecar injection is added automatically by a `ytt` [overlay](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/networking.yml#L63).
+
+To confirm your namespace has automatic sidecar injection enabled, you can describe your namespace with `kubectl get namespace <namespace_name> -L istio-injection` and see that Istio Injection is set to "enabled".
+
```
heycait, line 121: It appears the istio components living in the `istio-system` namespace have a metrics endpoint at ":15014/metrics". However, we intend to deploy Prometheus into the `cf-system` namespace which has the sidecar injection enabled, whereas `istio-system` does not. How would we be able to scrape the istio metrics?

config/network-policies.yaml
```diff
@@ -0,0 +1,563 @@
...
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
+    - podSelector:
+        matchLabels:
+          app.kubernetes.io/name: cf-api-server
+      namespaceSelector:
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
+---
+kind: NetworkPolicy
```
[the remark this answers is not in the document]
tcdowney: You could give a permissive policy to the `prometheus` -- allow it to egress to anything -- and then for the components that have metrics endpoints they could either:

1. Write an ingress policy to explicitly allow prometheus (this would end up in a lot of entries probably) for each component
2. We could coordinate and have system components serve their metrics scrape endpoints on a specific well-known port and have a blanket allow ingress rule for that port
3. Assuming you label prometheus with a well-known label, have a blanket allow ingress rule for the namespace that says that pods should allow ingress from something with that label
3. Some mixture of the above?

docs/component_authors/networking.md
tcdowney: I don't think it should be a problem. Both Pods must have sidecars if the "server" Pod has an Istio sidecar. The client Pod should be able to talk to other things that don't have one (based on how we've configured it at least).

[client sidecar] ----------any protocol---------> [something *without* sidecar] ✅
[client sidecar] ---------mTLS----------> [something with sidecar] ✅
[client *without* sidecar] ---------any protocol----------> [something with sidecar] 🛑

tcdowney: We'll try and take another pass at this section of the docs to clarify this.

heycait: So you're saying because istio components don't require it, the Prometheus pod should be able to connect regardless?

I added the network policy below and added annotations to istio-telemetry but am getting this error when Prometheus tries to scrape:  `server returned HTTP status 503 Service Unavailable`.

```
---
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: allow-prometheus-scrape
  namespace: istio-system
spec:
  podSelector: {}
  policyTypes:
  - Ingress
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
      podSelector:
        matchLabels:
          what-am-i: prometheus
```

config/network-policies.yaml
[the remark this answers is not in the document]
heycait: Options 2/3 seem like good options with 3 probably being the most ideal. Duplicating ingress across each component would be very unfortunate.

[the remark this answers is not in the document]
heycait: What would Option 3 policy look like?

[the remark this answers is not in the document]
kkburr: I like option 3 as well. Regarding 2, I think CAPI needed to use 9102 for statsd_exporter.

I really like the loop idea and feel like could make implementing a netpol for prometheus simple if we are leaning towards option 1. I might even vote to put prometheus in its own namespace since it doesn't seem like we're gaining anything by putting it in cf-system anymore.

docs/component_authors/networking.md
kkburr: Do you also have to add a network policy that allows egress from the prometheus pod to that ns?

config/network-policies.yaml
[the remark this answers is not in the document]
tcdowney: Option 3 would look similar to this policy that allows anything in `istio-system` to talk to all of the Pods in `cf-system`:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/files#diff-652a3e228db9cbb604d4ba5801be220fR47

A similar policy could be made that matches to `prometheus`.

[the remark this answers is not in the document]
tcdowney: > I might even vote to put prometheus in its own namespace since it doesn't seem like we're gaining anything by putting it in cf-system anymore.

I think `cf-for-k8s` as a project should have a larger discussion on this. I think there are some operational benefits to consolidating "cf related stuff" in `cf-system`, but am curious to hear others' thoughts here and would love project-wide guidance on the subject. 😅

cc @jenspinney @jamespollard8 @davewalter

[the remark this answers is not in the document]
jamespollard8: I'm inclined to lean into "do what works" here (as opposed to following rules about consistently consolidating or separating components into namespaces).

So, maybe that means put things in cf-system by default AND have no qualms about putting prometheus in its own namespace if that helps us operationally/maintenance-wise.

I'm definitely interested in hearing from others (eg Jen, Dave, etc)

shamus:
Here's the @cloudfoundry/cf-uaa team's 2 cents:

The UAA team doesn't know which components will want to talk to the UAA (whether to retrieve a token, validate it, or take some other action). That is to say, the UAA is used in a general purpose way, like the database. For this reason we don’t know that it makes sense for the UAA to be the shepherd of this NetworkPolicy.

Looking [here](https://github.com/cloudfoundry/cf-for-k8s/blob/master/config/uaa.yml#L110) we see five clients defined, but it isn't clear if those clients are used by the components identified by this networking policy or not.

The policy in this PR is very explicit; it will need to be updated anytime a new component needs access to the UAA.

Another model that would work is a hub / spoke like policy:

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
 name: allow-uaa-access
spec:
 podSelector:
  matchLabels:
   app: "uaa"
 policyTypes:
 - Ingress
 ingress:
 - from:
  - podSelector:
    matchLabels:
     networking/allow-uaa-access: "true"
```

Which would allow components to opt into being able to reach the UAA.

Which approach is better really depends on the security goals for CF. The benefit of the approach in this PR is its explicitness. The benefit of the hub / spoke approach is its flexibility.

config/network-policies.yaml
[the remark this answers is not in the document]
heycait: > Option 3 would look similar to this policy that allows anything in `istio-system` to talk to all of the Pods in `cf-system`:
> https://github.com/cloudfoundry/cf-for-k8s/pull/270/files#diff-652a3e228db9cbb604d4ba5801be220fR47
> 
> A similar policy could be made that matches to `prometheus`.

I see, and we'd need one of these for all namespaces then. 

I'd also like guidance about what namespaces we should be using as well.

docs/component_authors/networking.md
heycait: Played around with this more and was actually able to see metrics for some of the istio-components using the above network policy. However only half of the istio components worked (citadel, pilot, sidecar-injector) and the other half fails (galley, telemetry, policy). I can open this as an issue directly on cf-k8s-networking though.

tcdowney: Hmm, a lot of those components are being consolidated as part of the Istio 1.6.x upgrade (we'll end up with just two processes `istiod` and `istio-ingressgateway`) so this might change a bit anyway... but I think if you could create the issue that would help us keep it on the radar!

tcdowney:
Thanks for the feedback and reviews, everyone! We acknowledge the friction this adds and are going to discuss it a bit more next week. 

Long way of saying, please don't merge this PR as is right now!

cc @shalako @keshav-pivotal

config/network-policies.yaml
tcdowney: @piyalibanerjee made this change:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/commits/1b0f40ef1ba6f7b7fc0b4b4a3d21ed8698930b9b

Added policy for any Cloud Controller component that has a `fog_connection` config.

docs/component_authors/networking.md
tcdowney: We attempted to make that part of the docs more clear. Let us know if you have any other suggestions!
https://github.com/cloudfoundry/cf-for-k8s/pull/270/commits/8d3fd6e6d2142f0f58855f5b89d509437d26cde1

config/network-policies.yaml
Benjamintf1: @tcdowney at some point(in the not so far future) we plan on making fluentd able to egress to anywhere specified off platform, but I think only to log-cache should be fine at the moment

Benjamintf1: Let me rephrase, I was remembering wrong. So, for fluentd. Fluentd needs to contact(at the moment), the kubernetes api, and also log-cache(fluentd initiates), and will in the future need to contact any user defined source(so contact all makes sense. If there was a way to unlimit external ip's, or for us to dynamically limit ips(might be very difficult and not worth it), maybe we could, but the way I see it unlimited egress from fluentd makes sense to me, then we could hypotheticially limit internal sources.

as far as ingress though, we'll need teams to be able to ingress into fluentd. Obviously unlimited would work, but if we want to limit it further that could work too?

config/network-policies.yaml
```diff
@@ -0,0 +1,581 @@
...
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
+  egress:
+  - to:
+    - podSelector:
+        matchLabels:
+          app.kubernetes.io/name: uaa
+      namespaceSelector:
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/cf-system-ns: ""
```
Benjamintf1, line 538: what's this label and why is it blank?

config/network-policies.yaml
```diff
@@ -0,0 +1,581 @@
...
+  - Ingress
+  - Egress
+  podSelector:
+    matchLabels:
+      app: log-cache
+  ingress:
+  - from:
+    - podSelector:
+        matchLabels:
+          istio: ingressgateway
```
Benjamintf1, line 515: Is this what allows it to be accessed from the external lb?

config/network-policies.yaml
```diff
@@ -0,0 +1,581 @@
...
+      namespaceSelector:
+        matchLabels:
+          cf-for-k8s.cloudfoundry.org/istio-system-ns: ""
+    - podSelector:
+        matchLabels:
+          app.kubernetes.io/name: cf-api-server
```
Benjamintf1, line 521: What does cf-api-server need to contact log-cache for? I thought the only usecase for this was handled by the metric proxy now, now log-cache? Is there some usecase I'm missing here?

config/network-policies.yaml
tcdowney: It's basically a label that's used as a "tag" where you don't really care about it's value. The `cf-system` namespace has this label:
https://github.com/cloudfoundry/cf-for-k8s/blob/a339593d222cec1481a4e2a35be4e6bab0be881a/config/system-namespace.yml#L9

Sso this is basically saying `log-cache` can talk to a `Pod` labeled `app.kubernetes.io/name=uaa` that is in a namespace labeled `cf-for-k8s.cloudfoundry.org/cf-system-ns`.

config/network-policies.yaml
tcdowney: Yep. The External lb will point to the `istio-ingressgateway` Pods which will route traffic to log-cache. From loge-cache's perspective it is receiving ingress traffic from a Pod with the `istio=ingressgateway` label in a namespace with the `cf-for-k8s.cloudfoundry.org/istio-system-ns` label.

config/network-policies.yaml
tcdowney: I only knew that it used to talk to log-cache in CF for VMs. Maybe it doesn't need to anymore. 🤷 

Looking at the CCNG config it probably doesn't:
https://github.com/cloudfoundry/capi-k8s-release/blob/4a69758bf636c736dfd178b857ec13b46ee398a6/templates/ccng-config.lib.yml#L87-L102

tcdowney: @piyalibanerjee @cwlbraa 

Can you confirm that CC doesn't talk directly to log-cache anymore?

config/network-policies.yaml
Benjamintf1: Thanks

georgethebeatle:
Hi all, 

We (Eirini team) added what was missing in the network policies. We validated them on a cluster and all looks good. We have left egress wide open since all our components are using the kube API.

tcdowney:
Thank you @georgethebeatle!

tcdowney:
@cloudfoundry/cf-release-integration 

@keshav-pivotal and @shalako have been talking with some of the other teams and getting alignment on this approach and it sounds like folks are on board now. We described the expected workflow in the [writing a NetworkPolicy section](https://github.com/cloudfoundry/cf-for-k8s/blob/7a9ee9571538c01534661291734c6f510647ff5a/docs/component_authors/networking.md#writing-network-policies) of the docs we added as part of this story.

Anyways, we believe this is ready to merge now! Thanks.

config/network-policies.yaml
davewalter: Is this still an open question? If not, could someone please update this thread and resolve it?

config/network-policies.yaml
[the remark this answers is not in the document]
davewalter: Is this still an open question? If not, could someone please update this thread and resolve it?

config/network-policies.yaml
davewalter: Is this still an open question? If not, could someone please update this thread and resolve it?

config/network-policies.yaml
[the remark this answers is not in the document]
tcdowney: I think the docs we wrote around [adding a NetworkPolicy](https://github.com/cloudfoundry/cf-for-k8s/blob/7a9ee9571538c01534661291734c6f510647ff5a/docs/component_authors/networking.md#writing-network-policies) for new components addresses this.

config/network-policies.yaml
tcdowney: Summarizing what was said:
Since `fluentd` calls out to the Kubernetes API it needs open egress here since we don't know if the API is hosted on cluster or external to the cluster. Additionally it sounds like `fluentd` Open egress here is correct, at least for now.

Since components are sending stuff directly to `fluentd` it sounds like we're missing some allow rules for `ingress` here.

I don't think we want `ingress` to be truly unlimited from all namespaces (I think it would be bad if apps can arbitrarily send stuff to log-cache... might be able to pollute the logs for other apps), so I'll add rules for the `cf-api-*` jobs and `istio-ingressgateway`.. I think those are the things that are or will soon be sending logs directly to `fluentd` via the network.

config/network-policies.yaml
cwlbraa: the section of config that tim pointed out is authoritative, afaik.

tcdowney:
@davewalter I moved it into `config/networking`:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/commits/ba622782ee2e6d099633e3bfa47825a14ab34ba6

config/network-policies.yaml
tcdowney: Removed those two `ingress` rules:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/commits/b9b42d035ae86440f5648bfaf4bdfb5355cde6e9

config/network-policies.yaml
tcdowney: Added these `ingress` rules:
https://github.com/cloudfoundry/cf-for-k8s/pull/270/commits/b9b42d035ae86440f5648bfaf4bdfb5355cde6e9

PR #271, opened 2020-07-06
smoke-test: Increase timeout of cf-push to 10 min

phil9909:
Especially the first `cf push` sometimes takes longer than 5 minutes, which is the current timeout.

### DIFF
--- a/tests/smoke/smoke_test.go
+++ b/tests/smoke/smoke_test.go
@@ -90,7 +90,7 @@ var _ = Describe("Smoke Tests", func() {
 
 			By("pushing an app and checking that the CF CLI command succeeds")
 			cfPush := cf.Cf("push", appName, "-p", "assets/test-node-app", "--no-route")
-			Eventually(cfPush).Should(Exit(0))
+			Eventually(cfPush, 10*time.Minute).Should(Exit(0))
 			mapRoute(appName)
 
 			By("querying the app")


jamespollard8:
Thanks @phil9909 - we're fascinated to hear about this timing data.

Could you please clarify what specifically you mean by "first push". Is it first push for an app or for a cluster?

Also, could you please share with us:
1) general details about what setup you're using that's giving such slow `cf push`es?
2) 2 sets of timing data. W'd be curious to see results from:

```
time cf push test-node-app1 -p <path to asset>
time cf push test-node-app1 -p <path to asset>
time cf push test-node-app2 -p <path to asset>
kapp delete -a cf -y
<re-kapp deploy your cf>
time cf push test-node-app1 -p <path to asset>
time cf push test-node-app1 -p <path to asset>
time cf push test-node-app2 -p <path to asset>
```

phil9909:
We install the cf using `kapp-controller` v0.6.0 on a 1.16.10 cluster provided by Gardener on GCP and cluster-autoscaling from 4 to 7 nodes.
While the first push was in progress we noticed a scale-up from 4 to 5 nodes. This is very likely the reason for the delay. I will repeat the measuring with autoscale settings 5 to 7 nodes.

|          |  duration |
|----------|----------:|
| 1st push | 6m24.162s |
| 2nd push | 1m26.249s |
| 3rd push |  1m0.826s |
| redeploy |
| 4th push | 1m38.027s |
| 5th push | 0m46.588s |
| 6th push |  1m13.462s |

Notice: redeploying the cf did not trigger a scale-down of the cluster, so after the first push until the very last one there where 5 nodes registered in the cluster.

phil9909:
|          |  duration |
|----------|----------:|
| 1st push | 2m35.866s |
| 2nd push | 0m57.854s |
| 3rd push |  1m52.446s |

By first push I mean the very first push to a (newly created) k8s-cluster.

jamespollard8:
Oh cool, that's great to know - thanks for running those tests.

Obviously the tradeoff we're looking at is faster failing feedback vs a more robust test.

Because our deploy documentation suggests a minimum of 5 nodes, I'm thinking the current 5 minute timeout is actually ideal if you're seeing 2.5 minutes for the 1st push on a properly sized cluster, and if the >5 minute runs are a quirk of the cluster.

What do you think @phil9909? Should we close this or would you still want to increase the timeout?

phil9909:
I guess it would be fine for us to change our default values. @loewenstein do you agree?

BTW: There is a related Slack-Thread https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1594228249338000

loewenstein:
I agree, although we should think about requiring a particular number of vCPU and RAM instead of number of fixed size workers.

jamespollard8:
> I agree, although we should think about requiring a particular number of vCPU and RAM instead of number of fixed size workers.

Sounds good. FWIW we do also include suggestions for VM size in that same doc.

PR #279, opened 2020-07-14
Fix for repeated secret creation by kapp-controller

lucaschimweg:
See https://github.com/k14s/kapp/issues/124 for details

### DIFF
--- a/config/0-kapp-serviceaccount-fix.yml
+++ b/config/0-kapp-serviceaccount-fix.yml
@@ -1,0 +1,8 @@
+apiVersion: kapp.k14s.io/v1alpha1
+kind: Config
+
+# Fix for https://github.com/k14s/kapp/issues/124
+diffAgainstLastAppliedFieldExclusionRules:
+- path: [secrets]
+  resourceMatchers:
+  - apiVersionKindMatcher: {apiVersion: v1, kind: ServiceAccount}
--- a/config/0-kapp-serviceaccount-fix.yml
+++ b/config/0-kapp-serviceaccount-fix.yml
@@ -1,7 +1,7 @@
 apiVersion: kapp.k14s.io/v1alpha1
 kind: Config
 
-# Fix for https://github.com/k14s/kapp/issues/124
+#! Fix for https://github.com/k14s/kapp/issues/124
 diffAgainstLastAppliedFieldExclusionRules:
 - path: [secrets]
   resourceMatchers:


cppforlife:
@lucaschimweg i dont think this is a viable solution as this would make kapp ignore any different to this service account's spec (in this case actually all ServiceAccounts). which means if cf-for-k8s wants to modify it at a later point in any way, change will not be be applied since this tells kapp to ignore it. the real answer is to figure out how to do: https://github.com/k14s/kapp/issues/93, but we havent come up elegant solutions yet.

ericpromislow:
Given @cppforlife's comments, and that this doesn't pass current unit tests, and is in fact too aggressive, we're rejecting this PR.  Please continue to work with the `k14s` team to arrive at a good solution.

kramerul:
Hi @cppforlife,
actually it's not the case that kapp will ignore any difference to service account's with this additional configuration. Modifying the cf-for-k8s installation by adding an addition secret ('test') to the service account still leads to an application of the change. We also wondered how this could happen. Is this a bug?

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cc-kpack-registry-service-account
  namespace: #@ data.values.staging_namespace
secrets:
  - name: cc-kpack-registry-auth-secret
  - name: test
```

```bash
Target cluster 'https://api.uli.istio.shoot.canary.k8s-hana.ondemand.com' (nodes: shoot--istio--uli-worker-vxkas-z1-68ccc7979c-sxm6j, 3+)

Changes

Namespace             Name                               Kind            Conds.  Age  Op      Wait to    Rs  Ri  
cf-workloads-staging  cc-kpack-registry-service-account  ServiceAccount  -       13h  update  reconcile  ok  -  

Op:      0 create, 0 delete, 1 update, 0 noop
Wait to: 1 reconcile, 0 delete, 0 noop

7:21:01AM: ---- applying 1 changes [0/1 done] ----
7:21:02AM: update serviceaccount/cc-kpack-registry-service-account (v1) namespace: cf-workloads-staging
7:21:02AM: ---- waiting on 1 changes [0/1 done] ----
7:21:02AM: ok: reconcile serviceaccount/cc-kpack-registry-service-account (v1) namespace: cf-workloads-staging
7:21:02AM: ---- applying complete [1/1 done] ----
7:21:02AM: ---- waiting complete [1/1 done] ----
```

kapp works with this additional configuration exactly the way we want it to be although this might not be intended.

PR #280, opened 2020-07-14
Add quarks secret

peterhaochen47:
Things we have not included in this PR but will follow up later:
- In this PR, the QuarksSecret image is still stored on Credhub team's test repository; once we gain access to the cloudfoundry dockerhub we will make another PR to switch to using cloudfoundry dockerhub.
- We have not figured out how to track updates on QuarksSecret helm chart; we will communicate that once we figured it out (currently, vendir.yml just uses chart on the latest commit as of 7/14/2020 of QuarksSecret).

### DIFF
--- a/build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
+++ b/build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
@@ -35,7 +35,7 @@ rules:
   - create
   - update
 
-
+{{- if .Values.applyCRD }}
 - apiGroups:
   - apiextensions.k8s.io
   resources:
@@ -44,6 +44,7 @@ rules:
   - create
   - get
   - update
+{{- end }}
 
 - apiGroups:
   - ""
@@ -69,6 +70,7 @@ rules:
   - secrets
   verbs:
   - create
+  - delete
   - get
   - list
   - update
--- a/build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
+++ b/build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
@@ -1,0 +1,103 @@
+{{- if .Values.global.rbac.create }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  creationTimestamp: null
+  name: {{ template "quarks-secret.fullname" . }}
+rules:
+- apiGroups:
+  - certificates.k8s.io
+  resources:
+  - certificatesigningrequests
+  verbs:
+  - approve
+  - create
+  - delete
+  - get
+  - list
+  - update
+  - watch
+
+- apiGroups:
+  - certificates.k8s.io
+  resources:
+  - signers
+  resourceNames:
+  - kubernetes.io/legacy-unknown
+  verbs:
+  - approve
+- apiGroups:
+  - certificates.k8s.io
+  resources:
+  - certificatesigningrequests/approval
+  verbs:
+  - create
+  - update
+
+
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - create
+  - get
+  - update
+
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  verbs:
+  - get
+  - list
+  - watch
+
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - watch
+
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
+  - update
+
+- apiGroups:
+  - quarks.cloudfoundry.org
+  resources:
+  - quarkssecrets
+  verbs:
+  - delete
+  - get
+  - list
+  - update
+  - watch
+
+- apiGroups:
+  - quarks.cloudfoundry.org
+  resources:
+  - quarkssecrets/status
+  verbs:
+  - update
+{{- end }}
--- a/build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
+++ b/build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
@@ -1,0 +1,55 @@
+## Default values for Quarks Secret Helm Chart.
+## This is a YAML-formatted file.
+## Declare variables to be passed into your templates.
+
+
+# applyCRD is a boolean to control the installation of CRD's.
+applyCRD: true
+
+# fullnameOverride overrides the release name
+fullnameOverride: ""
+
+# image is the docker image of quarks secret.
+image:
+  # repository that provides the operator docker image.
+  repository: quarks-secret
+  # organization that provides the operator docker image.
+  org: cfcontainerization
+  # tag of the operator docker image
+  tag: foobar
+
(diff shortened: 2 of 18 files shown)


peterhaochen47:
Here are some example yamls:
```
---
apiVersion: v1
kind: Namespace
metadata:
  name: cf-system
  labels:
    istio-injection: enabled
    quarks.cloudfoundry.org/monitored: cf # add this label to tell Quarks to operate in this Namespace
---
apiVersion: quarks.cloudfoundry.org/v1alpha1
kind: QuarksSecret
metadata:
  name: quark-foo-secret
  namespace: cf-system 
spec:
  type: password
  secretName: foo-secret-1
```

build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
```diff
@@ -0,0 +1,103 @@
...
+- apiGroups:
+  - certificates.k8s.io
+  resources:
+  - certificatesigningrequests/approval
+  verbs:
+  - create
+  - update
+
+
+- apiGroups:
```
acosta11, line 39: Could we make this permission conditional based on the applyCRD value as well? It seems like the operator-based CRD creation flow is the only motivation for the permission. It worked on our cluster when we removed the permission and redeployed from scratch with the CRD resource included in the rendered yaml.

build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
```diff
@@ -0,0 +1,55 @@
+## Default values for Quarks Secret Helm Chart.
+## This is a YAML-formatted file.
+## Declare variables to be passed into your templates.
+
+
+# applyCRD is a boolean to control the installation of CRD's.
+applyCRD: true
```
acosta11, line 7: We would like to add a yaml file into the helm chart itself (or into the build dir if we must) for the CRD resource which is also conditionally rendered when we set applyCRD: false because
- We prefer to have least privilege for the operator
- We would like kapp to own all of the resources we are deploying, so it needs to include the explicit CRD yaml in its underlying kubectl apply
- It's easier to manage the creation of QuarksSecrets themselves if we are also creating the CRD on which they are dependent (we can order these in kapp, for instance)

As a side note it was a bit weird that the authoritative CRD config is inlined json in the go code for the operator as it makes it harder for consumers to find. We ended up taking the result of the operator from our own kubernetes cluster to produce a standalone yaml file for the CRD.

That file is included below as txt, but is yaml formatted.
[quarkssecret-crd.txt](https://github.com/cloudfoundry/cf-for-k8s/files/4921948/quarkssecret-crd.txt)

build/quarks-secret/quarks-values.yaml
```diff
@@ -0,0 +1,9 @@
+fullnameOverride: cf-quarks-secret
+
+image:
+  repository: quarks-secret
+  org: us.gcr.io/cf-security-credhub-main
+  tag: latest
+
+global:
+  monitoredID: cf
```
acosta11, line 9: Do we need a different default here than the release fullname? We think it's fine to let the monitoredId be an arbitrary value at the helm template step so that we can consolidate the monitoredID configuration and namespace labelling all in a focused ytt overlay file.

tomkennedy513: ya that should be fine, we'll make that change

tomkennedy513:
We reached out to quarks-dev about some of your helm chart questions. https://cloudfoundry.slack.com/archives/C1BQKKNP4/p1594847496221700

build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
manno: There is an (outdated?) dump of the CRD in docs/crds, we had them in the file system originally, but having the CRD in code made CI and such a lot easier.

build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
manno: Ah, very nice. :+1:

build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
loewenstein: Helm's `crds` folder will not be taken into account for updates. I suppose this is not relevant here as the build script only uses `helm template`, but it would not work for regular `helm update - i` based workflows.

acosta11: > but having the CRD in code made CI and such a lot easier.

Makes sense. I imagine it would have been a lot of boilerplate to update if it had to live outside of code.

> Helm's crds folder will not be taken into account for updates.
Oh I didn't know Helm had conventions around crd resources, I'll do some reading. Thanks for the reminder that while cf-for-k8s is the current context of the integration, the quarks secret repo itself has a desire to properly support the helm chart and we should keep that tension in mind when making requests to update the chart.

adobley:
We [opened a PR](https://github.com/cloudfoundry-incubator/quarks-secret/pull/17) against quarks-secret to make some of the changes we are talking about.

When that is merged we will update the vendir to have the updated helm charts.

build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
```diff
@@ -0,0 +1,55 @@
...
+  rbac:
+    # create is a boolean to control the installation of rbac resources.
+    create: true
+  image:
+    # pullPolicy defines the policy used for pulling docker images.
+    pullPolicy: IfNotPresent
+  # monitoredID is a string that has to match the content of the 'monitored' label in each monitored namespace.
+  # The monitoredID helper uses the release fullname, unless this is set.
+  monitoredID:
+  rbac:
```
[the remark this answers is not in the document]
peterhaochen47: pulled in a new chart where this issue is resolved.

build/quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml
peterhaochen47: pulled in a new chart where this issue is resolved.

build/quarks-secret/_vendir/deploy/helm/quarks-secret/values.yaml
peterhaochen47: pulled in a new chart where this issue is resolved.

peterhaochen47:
@acosta11  We pulled in the latest helm chart, which should resolve the issues you raised.

peterhaochen47:
Quarks helm charts are also published on helm hub from now on https://hub.helm.sh/charts/quarks/quarks-secret

davewalter:
Hi @peterhaochen47,

It looks like the failing check comes from CI running the `quarks-secret/build.sh` script, which results in changes to the rendered.yml file. I confirmed that I see the same set of changes when I clone your fork and run the script manually. I think that the image reference changes come from the reference to `latest` in the `quarks-secret/quark-values.yml` file, which will probably need to reference the appropriate image tag to match the version being included (assuming that the images are appropriately tagged). There is also an addition to a list of verbs that seems to come from `quarks-secret/_vendir/deploy/helm/quarks-secret/templates/cluster-role.yaml`.

Regarding your question about automation, I think we are mostly good, apart from the image tag in the values file, which will require some special handling.

Hope this helps.
Dave

peterhaochen47:
New conflicts came up, likely because `origin/develop` has newer commits. Should we keep resolving these conflicts (until this PR is merged)?

paulcwarren:
We resolved the conflicts from the config refactor and created PR #310 

Thanks for the contribution

Issue #255, opened 2020-06-23
intermittent issue: app staging hangs after "Build successful" message


jamespollard8:
## Describe the bug
When `cf push`ing an app, we expect to see a
`Build successful` message shortly followed by a
`Waiting for app to start...` section.

Occasionally (eg these back-to-back builds [#3921](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/jobs/long-lived-env-sli/builds/3921) and [#3922](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/jobs/long-lived-env-sli/builds/3922)) we've see the push hang after the "Build successful" message and then timeout.

We've not yet looked carefully into this or figured out how to reproduce it. For now, we're opening this as a Known Issue.

jamespollard8:
Back in June, I think we were seeing this happen about 2-10% of the time.

Starting around Thursday July 9th, we started seeing this a lot more - sometimes 5 times in a row for certain pipelines.

We've been talking with the CAPI team about this on slack [here](https://cloudfoundry.slack.com/archives/C07C04W4Q/p1594421054306600?thread_ts=1594421052.306500&cid=C07C04W4Q) and plan to post higher-level updates about this on CF slack in #cf-for-k8s [here](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1594759616405800)

As a temporary patch, we've put retry attempts in a few of our pipelines (890cb749a60019c885eb7944561cccb2afc9aea8 and 978c4f3145b7103d00423e599a8c43f0b30dd186)

jamespollard8:
Fixed by https://www.pivotaltracker.com/n/projects/966314/stories/173823389

PR #282, opened 2020-07-14
FIX: revert CAPI back to 8d1cae7 from Tues 7/7

jamespollard8:
We're trying this out to see if this reduces the frequency of the cf push timeout flake captured by #255 

Co-authored-by: Paul Warren <pwarren@pivotal.io>


jamespollard8:
We ran 5 trails:
3 pushed successfully on the first try for `push-test-app` and for uptimer
1 pushed successfully on the first try for `push-test-app` but uptimer failed to push its app
1 failed its first **two** attempts of `push-test-app`. uptimer pushed just fine

Results summary:
60% perfect
80% success with some simple pipeline retry logic

PR #286, opened 2020-07-16
Remove example with `helm install` based in-cluster database

loewenstein:
As discussed with @Syerram, there is no use case for a manually deployed in-cluster but outside the Istio mesh database, hence we should remove the example.


davewalter:
Hi @loewenstein,

Thanks for this contribution. Could you please update the base branch to `develop`?

Thanks,
Dave

loewenstein:
@davewalter done.

modulo11:
Please close in favour of #298.

PR #291, opened 2020-07-21
Add external db check on pr to ci pipeline

lucaschimweg:
Add a check for external DB functionality to pull request checks in the concourse pipeline (#168)

We could not test this on our concourse. Integrations and configurations for GCP are missing.

*Acceptance Steps**
Update Concourse Pipeline and see if new pull request steps are working

@kramerul

### DIFF
--- a/ci/pipelines/cf-for-k8s.yml
+++ b/ci/pipelines/cf-for-k8s.yml
@@ -502,76 +502,13 @@ jobs:
     params: {acquire: true}
 
   - task: install-postgres
+    params:
+      GCP_PROJECT_NAME: ((ci_k8s_gcp_project_name))
+      GCP_PROJECT_ZONE: ((ci_k8s_gcp_project_zone))
+      GCP_SERVICE_ACCOUNT_JSON: ((ci_k8s_gcp_service_account_json))
+    file: cf-for-k8s-ci/ci/tasks/install-postgres/task.yml
     input_mapping:
       pool-lock: ready-pool
-    config:
-      <<: *common-task-config
-      inputs:
-      - name: cf-for-k8s-develop
-      - name: cf-for-k8s-ci
-      - name: pool-lock
-      outputs:
-      - name: db-metadata
-      params:
-        GCP_SERVICE_ACCOUNT_JSON: ((ci_k8s_gcp_service_account_json))
-        GCP_PROJECT_NAME: ((ci_k8s_gcp_project_name))
-        GCP_PROJECT_ZONE: ((ci_k8s_gcp_project_zone))
-      run:
-        path: /bin/bash
-        args:
-        - -ec
-        - |
-          source cf-for-k8s-ci/ci/helpers/gke.sh
-
-          cluster_name="$(cat pool-lock/name)"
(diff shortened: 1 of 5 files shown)


ci/pipelines/cf-for-k8s.yml
```diff
@@ -1442,6 +1443,235 @@ jobs:
...
+        APP_REGISTRY_REPOSITORY_PREFIX: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_USERNAME: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_PASSWORD: ((cf_for_k8s_private_dockerhub.password))
+        USE_EXTERNAL_DB: "true"
+      file: cf-for-k8s-ci/ci/tasks/install-cf-on-gke/task.yml
+      input_mapping:
+        cf-for-k8s: cf-for-k8s-develop
+
+  - task: push-test-app
+    attempts: 3
```
jamespollard8, line 1592: please remove these, they're no longer desirable

lucaschimweg: Which ones exactly?

lucaschimweg:
Refactored, both `install-postgres` tasks use the same task definition now

ci/pipelines/cf-for-k8s.yml
loewenstein: I suppose @jamespollard8  is referring to the multiple 'attempts'?

ci/tasks/install-postgres/task.sh
```diff
@@ -0,0 +1,72 @@
...
+EOT
+
+export POSTGRES_PASSWORD=$(kubectl get secret -n external-db postgresql -o jsonpath="{.data.postgresql-password}" | base64 -d)
+kubectl exec -n external-db statefulset/postgresql-postgresql -i -- psql "postgresql://postgres:$POSTGRES_PASSWORD@localhost:5432/postgres" < /tmp/setup_db.sql
+
+cat > db-metadata/db-values.yaml <<EOT
+#@data/values
+---
+capi:
+database:
```
kramerul, line 56: The indentation is not correct here.

ci/pipelines/cf-for-k8s.yml
jamespollard8: Yes, the multiple attempts have been removed since the PR was initially opened. 👍

ericpromislow:
We had to revert this commit because `set-pipeline` failed (which you can't run directly, unfortunately):
```
$ fly -t relint-ci set-pipeline -c ci/pipelines/cf-for-k8s.yml -p cf-for-k8s
error: yaml: unknown anchor 'on_failure_uptime_pr_comment' referenced
```

Between submission and merging, the `on_failure_uptime_pr_comment` anchor was replaced with the underlying inline code. I know, but the`context` field kept varying, so the anchored block was removed.

It would be easiest if you could create a new PR based on this one and that change. And this time we'll test it locally before pressing the big "Merge" button.

Thanks,
Eric

kramerul:
Hi @ericpromislow,

I openend a new PR: #328

PR #299, opened 2020-07-22
Move to better ytt pattern for encoded overlays

ewrenn8:
Better pattern for applying overlays to encoded yaml

Private slack convo: https://vmware.slack.com/archives/CUK1X0Z8Q/p1595441763103900


mike1808:
Does this work because of the change of `YAMLFragment` type becoming a regular YAML type?

ewrenn8:
@mike1808 yes, it's now essentially [this example](https://get-ytt.io/#example:example-overlay) from the playground.

jamespollard8:
Thanks @ewrenn8 for showing us via PR! I've shared this with my teammates to make sure they get their 👀  on it.

PR #313, opened 2020-07-24
rename kpack-watcher-client-secret to cf-api-controllers-client-secret

cwlbraa:
We'd like to rename the capi-kpack-watcher deployment to "cf-api-controllers" to match some new source code organization we've done in capi-k8s-release. In cf-for-k8s, the main thing that still bears that name is the client secret. 

This PR does change one left-hand-side key in `values.yml`, and also updates hack/generate-values to match.

- we are doing this rename in accordance with https://github.com/cloudfoundry/capi-k8s-release/blob/rename-kpack-watcher-to-controllers/decisions/0003-cf-api-controller-repo.md

**PLZ NOTE:**

This PR uses a branch of capi-k8s-release that has the full rename. When this is merged, we'll need to merge that branch or else automated bumps of capi-k8s-release will cause cf-for-k8s pipeline breakage. Can we rely on cf-for-k8s autobump automation magic to change the vendir back to aiming at capi-k8s-release/ci-passed after everything's merged?

That branch is here: https://github.com/cloudfoundry/capi-k8s-release/pull/59 and is referenced in this PR's vendir.yml. 

**Acceptance Steps**

look in the k8s api and see that there is no longer a kpack_watcher_secret, and the uaa config has no mention of it.

@ericpromislow

### DIFF
--- a/vendir.yml
+++ b/vendir.yml
@@ -8,7 +8,7 @@ directories:
   - path: .
     git:
       url: https://github.com/cloudfoundry/capi-k8s-release
-      ref: bb40b9d67db1c3d150791d773752d1d788cdef91
+      ref: rename-kpack-watcher-to-controllers
     includePaths:
     - templates/**/*
     - values/**/*
--- a/config/capi/_ytt_lib/capi-k8s-release/templates/controllers_deployment.yml
+++ b/config/capi/_ytt_lib/capi-k8s-release/templates/controllers_deployment.yml
@@ -3,11 +3,11 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: cf-api-kpack-watcher
+  name: cf-api-controllers
   namespace: #@ data.values.system_namespace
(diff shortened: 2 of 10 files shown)


vendir.yml
```diff
@@ -8,7 +8,7 @@ directories:
   - path: .
     git:
       url: https://github.com/cloudfoundry/capi-k8s-release
-      ref: bb40b9d67db1c3d150791d773752d1d788cdef91
+      ref: rename-kpack-watcher-to-controllers
```
jamespollard8, line 11: @cwlbraa @ericpromislow FYI using a branch causes issues for us. `vendir sync` gets out of date as the HEAD of the branch changes. I've switched to using the sha directly `6645721`.

build failure: https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/jobs/test-vendir-sync-on-cf-for-k8s/builds/269

PR #315, opened 2020-07-24
Exploration into using ginkgo to test drive component templates

paulcwarren:
`tests/ytt/external-db-enabled_test.go`

NB: the name of this test is clearly not correct as far as the test stands atm.  the point is we can run ytt templates and make assertions on the resultant yaml. 

Feedback Welcome


tests/ytt/external_db_test.go
```diff
@@ -0,0 +1,69 @@
...
+			}
+		})
+
+		It("should not have cf-db* namespace/statefulset", func() {
+
+			Expect(ctx).To(Not(ProduceYAML(
+				And(
+					WithNamespace("cf-db"),
+					WithStatefulSet("cf-db-postgresql").WithNamespace("cf-db")),
+			)))
```
davewalter, line 65: Would the "negative" testing interface make more sense if it was something like this?

```
Expect(ctx).To(ProduceYAML(
  And(
    WithoutNamespace("cf-db"),
    WithoutStatefulSet("cf-db-postgres"),
  )
))
```

tests/ytt/external_db_test.go
```diff
@@ -0,0 +1,69 @@
...
+				"uaa.database.host":      "",
+			}
+		})
+
+		It("should have cf-db namespace and cf-db-postgresql statefulset", func() {
+
+			Expect(ctx).To(ProduceYAML(
+				And(
+					WithNamespace("cf-db"),
+					WithStatefulSet("cf-db-postgresql").WithNamespace("cf-db")),
```
davewalter, line 42: Is there ever a case where you would want to assert on a StatefulSet and not assert on its namespace? I understand the composability of Gomega matchers, but I feel that it complicates the readability of the test.

(This comment makes more sense in the context of my comment about "negative" tests below.)

paulcwarren: Maybe not?  In general it seems things are always put into a namespace even if it is "default".  So perhaps it makes  sense to expand the signature to `WithStatefulSet(name, namespace string)` and just be explicit about it.

tests/ytt/external_db_test.go
paulcwarren: I can see this is more readable/understandable but its at the engineering cost of one matcher (not matcher) versus 2x the constructor functions.

I dont have a strong opinion either way.

Birdrock: It's probably more clear, but will increase the interface quite a bit. It would give more granular control - for example:
```
Expect(ctx).To(ProduceYAML(
  And(
    WithNamespace("cf-db"),
    WithStatefulSet("cf-db-postgres").WithoutNamespace("cf-db")),
  )
))
```
It'd be great material for refactor/getting community engagement. We could mark this material as good first issues.

davewalter: To avoid duplicating checks, can we write a matcher that is simply the negation of another matcher? At least that way we don't have to maintain both matchers and avoid them getting out of sync.

tests/ytt/external_db_test.go
Birdrock: Could we do a matcher that nests inside of ProduceYaml to specify namespace before any other matchers? Will we ever be asserting across namespaces?

tests/ytt/external_db_test.go
paulcwarren: @jamespollard8 and I decided to use the standard gomega not matcher in the end.

PR #322, opened 2020-07-28
update metric-proxy to optional tls release

louis-brann:
- use `githubRelease` for vendir
- flatten `config` directory structure

Signed-off-by: Caitlyn Yu <caitlyny@vmware.com>

### Context
This is an intermediate non-breaking change to remove TLS from metric-proxy. This is necessary because of CAPI dependencies.

It will be followed by an update to CAPI to remove TLS on the client side and then another update to metric-proxy to completely remove TLS. We'll make another PR when we make that new release.

- Slack conversations:
    - https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1594336433360700
    - https://cloudfoundry.slack.com/archives/CUW93AF3M/p1594666589255400

**Acceptance Steps**
- `metric-proxy` should be running and container metrics should still be emitted

pair: @heycait


jamespollard8:
In preparation for this, I updated our update-k8s-releases pipeline to use a github resource for metric-proxy. 46ab6fa

PR #245, opened 2020-06-17
Do not install cf-db if external databases are used for capi and uaa

c0d1ngm0nk3y:
> Thanks for contributing to cf-for-k8s!
>
> We've designed this PR template to speed up the PR review and merge process - please use it.

> _Please describe the change here._ 

If external databases are used for capi and uaa, there is no need to install an internal one. Only if at least one is not using `hostname` to specify a database host, we install an internal one and use its hostname.

- Make sure this PR is based off the `develop` branch
- Checkout the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/docs/contributing.md)
- Include any links to related PRs, issues, stories, slack discussions, etc... that will help establish context.

For documentation on external databases, see
https://github.com/cloudfoundry/cf-for-k8s/pull/143

- Is there anything else of note that the reviewers should know about this change?

**Acceptance Steps**

_please provide a series of instructions (eg kubectl or cf cli commands) for how our Product Manager can verify that your changes were properly integrated_

1. If you use an external database for capi **AND** uaa (by providing a hostname), then we should see no internal database and kapp deploy and smoketests should pass

2. If you just use defaults (use an internal database for either capi, uaa or both), then we should see the internal database and kapp deploy and smoketests should pass.

_Tag your pair, your PM, and/or team_

> _It's helpful to tag a few other folks on your team or your team alias in case we need to follow up later._

PR #326, opened 2020-07-29
Add optional templates to deploy external-dns for Google Cloud DNS

jspawar:
Added optional templates which deploys the [external-dns](https://github.com/kubernetes-sigs/external-dns) controller which will manage configured Google Cloud DNS entries. Reason being is so users deploying to GKE and also maintaining their DNS with Google Cloud DNS do not have to constantly, manually update DNS records when deploying or be forced to use terraform (with the templates in the `deploy` directory) and maintain terraform state.

Given that this requires adding new values to `config/values.yml`, would also like to open a discussion about how best to add these new, optional values. Is the way it is in the PR currently fine? Would you all prefer something along the lines of #245 with the `cfdb_enabled` function? 

We're a little unclear about how to proceed with this specifically given that it's an optional template (i.e. it's in `config-optional` and not the main `config` directory) whose values are implicitly optional as a result.

Also would love some feedback about us making IaaS-specific directories inside the `config-optional` directory like we did for this PR.

---

**Acceptance Steps**

1. Deploy cf-for-k8s with the following additional ytt templates:
    - `config-optional/use-external-dns-for-wildcard.yml`
    - `config-optional/gcp/add-external-dns.yml`
1. Verify that eventually the `*.<system_domain>` DNS record has been made in the desired Google Cloud DNS project
1. Ensure that cf has deployed successfully (can target the deployment with `cf api` or something which also further indicates DNS has been configured automatically)

Thanks,
Jwal + @matt-royal

### DIFF
--- a/config-optional/gcp/add-external-dns.yml
+++ b/config-optional/gcp/add-external-dns.yml
@@ -1,0 +1,86 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:base64","base64")
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: external-dns
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: external-dns
+  namespace: external-dns
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  name: external-dns
+rules:
+- apiGroups: [""]
+  resources: ["services","endpoints","pods"]
+  verbs: ["get","watch","list"]
+- apiGroups: [""]
+  resources: ["nodes"]
+  verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: external-dns-viewer
+  namespace: external-dns
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: external-dns
+subjects:
+- kind: ServiceAccount
+  name: external-dns
+  namespace: external-dns
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: google-cloud-dns-key
+  namespace: external-dns
+data:
+  key.json: #@ base64.encode(data.values.external_dns.gcp.service_account_key)
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: external-dns
+  namespace: external-dns
+spec:
+  strategy:
+    type: Recreate
+  selector:
+    matchLabels:
+      app: external-dns
+  template:
+    metadata:
+      labels:
+        app: external-dns
+    spec:
+      serviceAccountName: external-dns
+      containers:
+      - name: external-dns
+        image: registry.opensource.zalan.do/teapot/external-dns:latest
+        args:
+        - --source=service
+        - #@ "--domain-filter=" + data.values.external_dns.parent_domain
+        - --provider=google
+        - #@ "--google-project=" + data.values.external_dns.gcp.project_id
+        - --policy=upsert-only #! would prevent ExternalDNS from deleting any records, omit to enable full synchronization
+        - --registry=txt
+        env:
+        - name: GOOGLE_APPLICATION_CREDENTIALS
+          value: /etc/google-cloud-dns/key.json
+        volumeMounts:
+        - name: google-cloud-dns-key
+          mountPath: /etc/google-cloud-dns
+          readOnly: true
+      volumes:
+      - name: google-cloud-dns-key
+        secret:
+          secretName: google-cloud-dns-key
--- a/config/values.yml
+++ b/config/values.yml
@@ -156,3 +156,13 @@ metric_proxy:
     secret_name: "metric-proxy-cert"
     crt: ""
     key: ""
+
+#! [Optional] Used for optional config to set up `external-dns` deployment
+external_dns:
+  #! The domain which is the parent domain of `system_domain`
+  parent_domain: ""
+  gcp:
+    #! Use this to specify the project which contains the desired Google Cloud DNS entries
+    project_id: ""
+    #! The JSON key for a service account which has permission to modify the desired GCP project's Google Cloud DNS entries
+    service_account_key: ""


aegershman:
forgive me for the unsolicited opinion, I'm not trying to rag on anything or be negative. this has more to do with the overall strategy of bundling optional auxiliary deployments within cf-for-k8s than bundling external-dns specifically; 

I feel the path of bundling optionals auxiliary deployments alongside cf-for-k8s would cause more frustration and drift than benefit with getting started. I feel an alternative strategy to serve folks starting out with cf-for-k8s who want a more streamlined experience could be more examples of using other deployments co-existing alongside cf-for-k8s on the same cluster-- and things to be mindful of, like whether other namespaces include an `istio-injection: disabled` label, that sort of thing-- but assume they'll be managed separately from the main cf-for-k8s deployment itself. for example, the organization I'm in is using `external-dns` as well, but we're running on aws route53 and assuming instance profiles.

apologies, maybe my comment here is better suited for a separate issue or something; like I said, not trying to be negative or anything, I'm just sharing this from the perspective of someone working to bring cf-for-k8s into our enterprise environments, and I feel that we'll have more success by providing examples of setting up deployments co-located on the same cluster as cf-for-k8s, rather than bundling them directly into cf-for-k8s releases. That's possibly a bigger 'strategy' vision I suppose 🤷‍♂️

I just wanted to share; didn't mean to muddle or be discouraging on this specific PR 👍

jspawar:
@aegershman input is always welcome, no need to apologize, thanks though :)

Would you mind clarifying a bit though? Are you suggesting deploying something like this, the optional external-dns controller, via a separate GitHub repo to provide that example? 

If so, I'm not opposed to exploring that some more (although perhaps a separate issue dedicated to a discussion for just that might be more productive?)

aegershman:
(most important paragraph is at the bottom.)

apologies for the late & a bit sloppy reply here, I'm out of town this weekend, poor timing on my part; basically I'm suggesting the default assumption be that folks will bring their own additional deployments if they see fit, the reason being that cffork8s as a product shouldn't account for and maintain combinations of optional deployments folks & organizations would have reason to co-deploy, even if it's done at their own risk as cf-for-k8s doesn't _officially_ support it. 

for example, at work we use `external-dns` in our cf-for-k8s envs too, but we use it on `aws` with instance role profiles for route53; even though much of the external-dns config is the same as it would be for `{gcp,azure,etc.}`, there are config variations in each which won't be captured in our vendoring in cf-for-k8s as well as by folks who specifically maintain external-dns.

so instead of bundling deployments into cf-for-k8s directly, put down the groundwork patterns for assuming that if teams want to use external-dns or prometheus or cert-manager or {x,y,z}, they could bring their own. if they need to integrate other deployments into cf-for-k8s to take advantage of them (such as `external-dns`), they can modify cf-for-k8s components (e.g. annotations) using `ytt` overlays. all of this done at their own _risk_, of course. but I feel it's reasonable to entrust that {operators, users, engineers, pick your term, etc.} who are operationally responsible for these k8s envs are learned to configure optional k8s co-deployments for their org's needs beyond what cf-for-k8s will vendor.

the alternative being some way to give suggestions and point out caveats, like pointing out cffork8s' use of istio will set up automatic sidecar injection on all pods in a namespace unless the namespace is labeled with `istio-injection: disabled`, or finding a way to give examples of vendoring and templating other projects like one of the prometheus-operator deployments, etc.

in conclusion, I'm sorry I didn't mean to clutter up your PR with this, this is the last comment I'll make on it and I'll drop it. I'm not trying to act like an authority on any of this, I'm just some guy™ spouting opinions, so releng feel free to put me in my place. I'm not trying to fundamentally disagree on this PR, it's more like... I feel that including more optional deployments, especially `config-optional/` as a directory, implies "everything in here is officially supported, we endorse this is as the way to <deploy thing>, all config here follows semantic versioning rules, and it's ready to go for production use"-- I feel there's room for something 'softer', like "here's what you _could_ do to deploy, but it's a suggestion for ideas for you to bring your own deployments to the table-- but we don't endorse it for production, there aren't guarantees, this is just what you could do (at your own risk) if you wanted to try composing other deployments together with cf-for-k8s" 👍

jspawar:
I think there's value in this discussion still, this probably won't be the last time this comes up. 

Our intention with this PR by placing it in `config-optional` is also in large part intended to indicate something like this is a suggestion of how to deploy `external-dns` with the intention that it is community maintained and not for production use. I agree that this might be potentially misleading as something "ready for production" as well as possibly not allowing room for operators to deploy `external-dns`/similar things on their own.  

I'm not exactly clear to be honest what the guidance for the `config-optional` directory is, but I just assumed it was along the lines of the `operations` or the `operations/experimental` directories in `cf-deployment` which contained community-maintained, optional configurations for a BOSH-based CF deployment. Perhaps we could also submit a `README` into `config-optional` to make that clear these are suggestions/not intended for production usage? 

I think I would defer to release integration folks at this point for more clarity and guidance. Thoughts @Syerram?

paulcwarren:
Hi @jspawar, 

Apologies for the tardy response times on this.  

As you know we removed `config-optional/` a while back but I think we folded some of this functionality into the `use_external_dns_for_wildcard` property (maybe?).  But, we don't support deploying the external dns controller.

In general we haven't provided any support for specific IaaS in the repo outside of our `ci` directory which is really only for our internal use and as a reference to consumers who may find it useful.  We also went a different direction with our CI too, creating pools of clusters with pre-defined DNS/IPs and feeding those IPs into deployments via the`istio_static_ip` property.

Thoughts?

loewenstein:
I could imagine that wrapping cf-for-k8s as ytt library in another repo, say cf-for-k8s-gke could be a good place to provide GKE specific config like this. Not sure if we find enough maintainers for such a repo in cloudfoundry or cloudfoundry-incubator though. Maybe cloudfoundry-community?

jspawar:
No problem @paulcwarren, I think @loewenstein's suggestion is probably a better direction to go with this notion of at least IaaS-specific config.

As for said suggestion, I am going to create a separate issue for this repo to discuss that plan. That seem good to you all?

paulcwarren:
This sounds like a good way forward to me @jspawar

ericpromislow:
We're going to close this PR, since we've zigged a bit while this PR was following the zag path. 

We can see a point where we open up a `cf-for-k8s-community` repo for storing add-ons on top of cf-for-k8s. For now we're focusing on the core technology.

PR #327, opened 2020-07-29
Add setting to redirect HTTP requests to HTTPS

XanderStrike:
This change adds a setting to the values file that allows users to tell the platform to redirect HTTP requests to the HTTPS endpoint.

**Acceptance Steps**

Given I have a deployment of cf-k8s with an app,
And I have set `gateway: {https_only: true}`
When I send an http request to that app
Then I can see that I am redirected to https
And I receive a successful response from my application https.

Given I have a deployment of cf-k8s with an app,
And I have set `gateway: {https_only: false}`
When I send an http request to that app
Then I receive a successful response from my application over http.

CC @mike1808 @ndhanushkodi

### DIFF
--- a/config/networking/external-routing.yml
+++ b/config/networking/external-routing.yml
@@ -37,7 +37,7 @@ metadata:
 
 #! this gateway is used to expose CF services in the cluster, externally
 #@ load("gateway.lib.yml", "gateway")
---- #@ gateway(data.values.system_domain, data.values.app_domains, data.values.system_namespace, data.values.workloads_namespace)
+--- #@ gateway(data.values.system_domain, data.values.app_domains, data.values.system_namespace, data.values.workloads_namespace, data.values.gateway.https_only)
 
 #! This will set overlay a static IP for the istio load balancer.
 #@overlay/match by=overlay.subset({"kind": "Service", "metadata":{"name": "istio-ingressgateway"}})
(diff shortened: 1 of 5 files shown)


mike1808:
Hey @cloudfoundry/cf-release-integration I noticed that [concourse-ci/upgrade-check-uptime](https://release-integration.ci.cf-app.com/builds/335884) failed. We didn't change anything in the default configuration, it's probably a flake. How can we rerun that test?

paulcwarren:
That looks like the smoke test app push failed to move on from staging.  There is a flake in this area so I would ignore this for now.

Syerram:
@XanderStrike Generally, we want connections to be encrypted. I recommend we hardcode this into ingress config as a default redirect to HTTPS unless there are specific use cases where the platform teams need it as plain text for the entire foundation.

XanderStrike:
Correct, that's the current behavior. This will redirect to HTTPS by default, unless the platform operator configures it to allow HTTP. If you'd like to make that setting harder to find I suppose we could remove it from the sample values? Part of acceptance on our side was to make it configurable though.

XanderStrike:
Ignore that last comment, I misread 😐 

We'll make it default to HTTPS redirection.

XanderStrike:
@Syerram done! Let us know if you have any more feedback.

XanderStrike:
Any more needed from us? I see this is still unscheduled.

jamespollard8:
> I see this is still unscheduled.

Glad to hear that label is useful. We used to keep our community section at the bottom of our Backlog so issues+PRs would pretty much always say scheduled. Now, the "scheduled" label should mean it's actually being worked on.

PR #333, opened 2020-08-03
doc: document router access logs for operators

tcdowney:
This is an updated and improved version of our [cf-k8s-networking access log docs](https://github.com/cloudfoundry/cf-k8s-networking/blob/98f43bf4f7d596b9901b1f045aa7c74fbc405a5c/doc/access-logs.md). The goal is to help an operator make sense of the fields in the access logs themselves as well as to help determine the source of any potential latency.

A followup story will be done to add documentation for app developers once these logs are available via `cf logs`. 

Networking story: [#173839341](https://www.pivotaltracker.com/story/show/173839341)

**Acceptance Steps**

Try out the the `kubectl` command in the docs and see if they help you understand! 🙃 If they don't that's valuable feedback.

**cc/**
@keshav-pivotal @shalako @XanderStrike @ndhanushkodi


davewalter:
Thanks @tcdowney. We know some of the tests were failing, but that had nothing to do with this PR so we went ahead and merged it.

PR #328, opened 2020-07-30
Add external db check on pr to ci pipeline (second try)

kramerul:
Add a check for external DB functionality to pull request checks in the concourse pipeline (#168)

We could not test this on our concourse. Integrations and configurations for GCP are missing.

Acceptance Steps*
Update Concourse Pipeline and see if new pull request steps are working

See also #291

@lucaschimweg @modulo11

PR #334, opened 2020-08-03
Revert "Initial NetworkPolicies for System Components"

davewalter:
Reverts cloudfoundry/cf-for-k8s#270

This is failing when using an external DB. We will revisit this change once we have cut a new release.


loewenstein:
I wonder why this showed up only after merging. Are the PR tests for external databases missing anything?
cc @phil9909

phil9909:
RCA:

Looking at the [PR-resource in concourse](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/resources/cf-for-k8s-pr-all-branches-and-forks) and searching for the PR #270 (the one that caused the problem):
image

You can see that the job "validate-pr-on-gke-external-db" did not run for this PR. The timestamp of the resource is from 2020-07-24 

The PR #328, which added the "validate-pr-on-gke-external-db" job was merged a week later. 

Conclusion: When PR #270 was merged on the 3rd of August it had a stale green check-mark for the tests.

PR #336, opened 2020-08-03
Re-apply "Initial NetworkPolicies for System Components"

davewalter:
This PR re-applies the changes from #270 that were temporarily reverted due to issues seen in CI.

* We saw that the external DB validation job [failed](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/jobs/validate-cf-for-k8s-gke-external-db/builds/37) due to the lack of a `cf-db` namespace
* We saw smoke-tests [failed](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s/jobs/validate-cf-for-k8s-gke/builds/362) due to a timeout after successfully staging

### DIFF
--- a/config/minio/minio.yml
+++ b/config/minio/minio.yml
@@ -18,6 +18,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: cf-blobstore
+  labels:
+    name: cf-blobstore
 
 #@overlay/match by=overlay.subset({"kind": "Secret", "metadata": {"name": "cf-blobstore-minio"}})
 ---
--- a/config/networking/network-policies.yaml
+++ b/config/networking/network-policies.yaml
@@ -18,7 +18,7 @@ to:
       release: cf-db
 #@ else:
 - ipBlock:
-    cidr: 0.0.0.0/32
+    cidr: 0.0.0.0/0
 #@ end
 ports:
 - protocol: TCP 
--- a/config/networking/network-policies.yaml
+++ b/config/networking/network-policies.yaml
@@ -3,6 +3,30 @@
 #! TODO: Should be labeled in capi-k8s-release
 #@ load("@ytt:overlay", "overlay")
 #@overlay/match by=overlay.and_op(overlay.subset({"metadata":{"name":"cf-workloads-staging"}}), overlay.subset({"kind": "Namespace"}))
+#@ def cfdb_enabled():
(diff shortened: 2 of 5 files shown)


kramerul:
Hi @davewalter,
we(me and @modulo11) had a look at the problem with the external database. We created a [branch|https://github.com/sap-contributions/cf-for-k8s/tree/network-policies-database] with the required changes.
With this changes, the external database scenario can be deployed, **but** we were not able to get it working. `ccdb-migrator` was waiting to get the configuration from pilot. We tried to modify the network policies but we didn't succeed.

Perhaps it would be useful to have separate files for the network policies for different namespaces. Currently the file is quite huge.

lisamburns:
> Hi @davewalter,
> we(me and @modulo11) had a look at the problem with the external database. We created a [branch](https://github.com/sap-contributions/cf-for-k8s/tree/network-policies-database) with the required changes.
> With this changes, the external database scenario can be deployed, **but** we were not able to get it working. `ccdb-migrator` was waiting to get the configuration from pilot. We tried to modify the network policies but we didn't succeed.
> 
> Perhaps it would be useful to have separate files for the network policies for different namespaces. Currently the file is quite huge.
> 
> CC @Syerram

We took a look at this and tried to deploy this with an AWS RDS on our own. We reproduced the error of not being able to connect. Then we changed the cidr on the network policies to `0.0.0.0/0' rather than '0.0.0.0/32'. After that it seemed to connect to the database, although there were other errors.
Context about the CIDR block in our conversation with networking team here:
https://cloudfoundry.slack.com/archives/CFX13JK7B/p1596666952261000

CC @acosta11

kramerul:
Hi @lisamburns,
yes the CIDR was wrong. Please use `0.0.0.0/0`.

acosta11:
With persistent failure feedback from CI, I gave this update a try locally and found:
- if you deploy develop first and then cf-api-clock is not rolled along with the network policy change, cf push continues to succeed
- if cf -api-clock is restarted after the network policy change, it hits a postgres connection error on first pod initialization, then does not log a postgres failure on subsequent restarts, but continues to fail with an upstream connection error?

I suspect we still have an error either in the cf-api-clock network policy or the cf-db network policy.

acosta11:
The failure in concourse-ci/newest-supported-k8s-version is a known issue in CI at the moment. Going ahead with this merge.

PR #348, opened 2020-08-05
Fix status context for external-db pipeline jobs

kramerul:
Currently the status of builds related to external database jobs are reported with the same context as other tests.
This PR will fix this.

This came up, because the reason for the failure of #336 was not clear.

We have no experience with the status reporting in github. We suppose that before merging this PR, you need to [create a new status for github](https://developer.github.com/v3/repos/statuses/).

**Acceptance Steps**

* Create status in github
* Merge PR
* Apply new pipeline

@modulo11 
CC: @Syerram

### DIFF
--- a/ci/pipelines/cf-for-k8s-contributions.yml
+++ b/ci/pipelines/cf-for-k8s-contributions.yml
@@ -478,7 +478,7 @@ jobs:
     params:
       path: cf-for-k8s-pr-all-branches-and-forks
       status: pending
-      context: upgrade-check-uptime
+      context: external-db
   - task: create-tf-vars-file
     config:
       platform: linux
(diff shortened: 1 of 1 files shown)


davewalter:
Thanks!

davewalter:
@kramerul @modulo11 

Just spotted that this PR only updated the pipeline definition for `cf-for-k8s-contributions` and not the template that this is generated from.

I updated the template to match this change [here](https://github.com/cloudfoundry/cf-for-k8s/commit/97ae90abbf8177b5f94f952c69284e7ac6e9466b). We also are planning to make it clearer that the pipeline itself is auto-generated to help avoid this issue in the future.

Thanks again for the fix!

PR #349, opened 2020-08-05
Bump Istio to 1.6.4

ndhanushkodi:
Bump Istio to 1.6.4

---

We've been blocked in the past on bumping Istio due to incompatibility with KinD. This PR makes the bump to Istio and the necessary optional configurations so that cf-for-k8s can be deployed with Istio 1.6.4 on KinD.

Story [#173382670](https://www.pivotaltracker.com/story/show/173382670)

**Acceptance Steps**
**Given** I have deployed cf-for-k8s.
**When** I describe an `istio-ingressgateway` pod with `kubectl -n istio-system get pods -l app=istio-ingressgateway -o yaml | grep image:`
**Then** I can see that the `istio-ingressgateway` pod uses the image `gcr.io/cf-routing/proxyv2:1.6.4`

@jenspinney @keshav-pivotal @shalako


jenspinney:
For the first two failures on the PR checks, neither us were able to get the concourse details to load. After we sign in to concourse, we just see "loading" indefinitely. Could someone help us out with the details of what failed in the "newest-supported" and "oldest-supported" jobs?

We aren't surprised to see that upgrade tests failed, since we know istio 1.4 cannot be upgraded to istio 1.6 and will require a full redeploy.

jamespollard8:
> For the first two failures on the PR checks, neither us were able to get the concourse details to load. After we sign in to concourse, we just see "loading" indefinitely. Could someone help us out with the details of what failed in the "newest-supported" and "oldest-supported" jobs?
> 
> We aren't surprised to see that upgrade tests failed, since we know istio 1.4 cannot be upgraded to istio 1.6 and will require a full redeploy.

sorry about that @jenspinney @ndhanushkodi , looks like we need to make sure all these concourse jobs are public.

Here's the error during the install: `ytt: Error: Checking file 'config-optional/remove-ingressgateway-service.yml': lstat config-optional/remove-ingressgateway-service.yml: no such file or directory`

jamespollard8:
Oh, and we'd expect to see CI issues because CI uses HEAD of develop - not your PR branch. I can help with this. (I'll temporarily have the task come from this branch.)

I think another pair has some pipeline changes in flight so I'll need to coordinate with them.

config/values.yml
```diff
@@ -156,3 +156,6 @@ metric_proxy:
     secret_name: "metric-proxy-cert"
     crt: ""
     key: ""
+
+opi:
+  unsafe_allow_automount_service_account_token: false
```
davewalter, line 161: Is this configuration used anywhere? I didn't see a reference to `data.values.opi.unsafe_allow_automount_service_account_token`.

jenspinney: You're right, this isn't used anywhere. This was from a different approach we thought about going down when we first started this work.

We'll remove this.

jenspinney: Removed

davewalter:
Note. This change is known to break upgrades, which is why we merged it despite failing CI checks.

jenspinney:
Thanks Dave!

PR #360, opened 2020-08-06
Deploy without LoadBalancer services by default

XanderStrike:
We want a default deployment of cf-for-k8s to function in an environments where LoadBalancer type services are not supported, such as Kind.

Also bumps cf-k8s-networking to version 0.3.0 which includes the following [Change Notes](https://github.com/cloudfoundry/cf-k8s-networking/releases/tag/untagged-ca7c7e9476fb718821e4)

### Acceptance Steps

As a platform operator deploying to a kubernetes environment that supports LoadBalancer
When I follow deploy.md to deploy my cluster
I realize that I need to include an optional overlay to utilize LoadBalancer
I can deploy a reachable cf-for-k8s

<hr>

As a platform operator deploying to a k8s environment that _does not_ support LoadBalancer
When I follow deploy.md to deploy my cluster
I realize that I do not need to include an optional overlay to utilize LoadBalancer
I can deploy a reachable cf-for-k8s

# This is a pretty significant breaking change to everyone's pipelines!!!1!

Actions required:

**If you're deploying to GKE** or anywhere else that supports LoadBalancer services, your `ytt` command will now need to **include** 
`-f config-optional/ingressgateway-service-loadbalancer.yml`.

**If you're deploying to Kind** or anywhere else that _does not_ support LoadBalancer services, you will need to **remove** the now-deleted `-f config-optional/ingressgateway-service-nodeport.yml` from your `ytt` command.

Special thanks to @ndhanushkodi @jenspinney


loewenstein:
What is the reason to default to environments such as KinD (without Load Balancer)?

davewalter:
Looks like our CI checks didn't trigger for some reason. Going to close and reopen this to see if that causes them to run.

jamespollard8:
> What is the reason to default to environments such as KinD (without Load Balancer)?

@loewenstein It sounds like our main reasoning is that most production environments will likely be running on environments without a LB (eg EKS and on-premise solutions like Vsphere).

loewenstein:
@jamespollard8 mmh, that sounds like a matter of perspective. I'd expect a lot of production environments running on cloud providers with load balancer. Could we hold this PR until the values vs overlay decision is made?
Maybe we also need a way to figure out which default the community as a whole would prefer.

Syerram:
@loewenstein 
> Could we hold this PR until the values vs overlay decision is made?
> Maybe we also need a way to figure out which default the community as a whole would prefer.

Please take a look at my comment in https://github.com/cloudfoundry/cf-for-k8s/pull/363.

davewalter:
We believe that this PR has been obsoleted by more recent changes to cf-for-k8s. If you still want to change the default for the `load_balancer.enable` flag from true to false, could you please open a new PR with the appropriate changes?

PR #363, opened 2020-08-07
Convert optional overlays to config flags

davewalter:
>

---

- If you're adding/removing/changing any values in `config/values.yml`, please review [this "cf-for-k8s values for contributors" doc](https://docs.google.com/document/d/1Y3jAx48TCGIQdzOFmzqp_R_0WT8Hfjx9oyqs2Tk0Otw/edit#) for up-to-date principles and guidelines for contributing values changes.
- Checkout the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/community/CONTRIBUTING.md)

**Acceptance Steps**

Deploy as normal. Use the appropriate data values flags instead of including optional overlays as desired/required.

@cloudfoundry/cf-release-integration


config/eirini/enable-automount-service-account-token-for-kind.yml
```diff
@@ -0,0 +1,27 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:yaml", "yaml")
+
+#@ if data.values.patch_for_kind:
```
jamespollard8, line 5: Cool - this makes sense and looks clean 👍🏽

config/metrics/add-metrics-server-components.yml
```diff
@@ -0,0 +1,156 @@
+#@ load("@ytt:data", "data")
+
+#@ if data.values.add_metrics_server:
```
jamespollard8, line 3: Nice. I think we also have some docs to update for this one.

config/metrics/patch-metrics-server-for-kind.yml
```diff
@@ -0,0 +1,23 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+
+#@ if data.values.patch_for_kind:
```
jamespollard8, line 4: I love that this `patch_for_kind` value helps apply multiple yml files - which I think is a big part of why we wanted to move towards this values API/interface.

config/networking/first-party-jwt-istio-for-kind.yml
```diff
@@ -0,0 +1,57 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:json", "json")
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:yaml", "yaml")
+
+#@ if data.values.patch_for_kind:
```
jamespollard8, line 6: Hmm, it sounds like this config-optional file was not as kind-specific as RelEng found that they need this for PKS/vsphere [vmware slack thread](https://vmware.slack.com/archives/C3LUD2L04/p1596764860097900?thread_ts=1596762799.097400&cid=C3LUD2L04)

jamespollard8: maybe that's solved by adding an `or` here?

config/networking/remove-resource-requirements.yml
```diff
@@ -0,0 +1,25 @@
+#@ load("@ytt:data", "data")
+#@ load("@ytt:overlay", "overlay")
+
+#@ if data.values.remove_resource_requirements:
```
jamespollard8, line 4: maybe add an `or data.values.patch_for_kind` ?

config/uaa/enable-automount-service-account-token-for-kind.yml
```diff
@@ -0,0 +1,13 @@
+#@ load("@ytt:data", "data")
```
jenspinney, line 1: For some historical context here (from my view, which may be missing lots of data and perspectives):

1. We (cf-k8s-networking) started the work to support Kind with Istio 1.6.4 a couple months ago. In our original design, we were planning to use data values to thread this info through. We had PRs and conversations with the eirini and uaa teams to add the necessary configuration to their subcomponents before we could tie it all together at the cf-for-k8s level.
2. For several reasons, the Istio 1.6.4 work went on longer than we expected. By the time we picked up this work again, I think the context of the original plan of using data values was lost. We instead went down the path of overlays at this point, but we had some vestigial pieces showing our original approach. @davewalter noticed one of these [while reviewing the PR](https://github.com/cloudfoundry/cf-for-k8s/pull/349#discussion_r466572563) that ultimately got Istio 1.6.4 into cf-for-k8s.
3. When we were recently preparing the PR for bumping Istio, we noticed that this change in direction had happened, probably unintentionally, but we wanted to first focus on getting Istio bumped at all before considering whether we wanted to move back to the data values approach.
4. Given all that, this file here is probably overly complicated compared to just passing a data value through when we generate the uaa library's data values. That was our original plan and the reason for [this PR](https://github.com/cloudfoundry/uaa/pull/1326) to UAA, and we're happy to help simplify some of this.

Sorry the verbosity for a relatively simple thing. I just wanted to have somewhere to capture what our intentions were and how we got here :).

config/metrics/patch-metrics-server-for-kind.yml
jenspinney: I don't have a strong opinion or deep knowledge here, but we may want to think about separating the automount-service-account-token thing from being 1:1 coupled with using Kind. Kind happens to be a Kubernetes provider that doesn't support [third party tokens](https://istio.io/latest/docs/ops/best-practices/security/#configure-third-party-service-account-tokens) right now, but:

1) There are probably other Kubernetes providers that also don't support this (likely including Minikube, but I haven't checked), and
2) Maybe someday Kind will start supporting this feature, so we wouldn't want to automount service account tokens there anymore, but we do need to apply other sorts of changes specific to Kind.

Just food for thought.

jenspinney: Oh, I see now something that might be related to this concern in a comment below :)

config/metrics/add-metrics-server-components.yml
davewalter: We didn't attempt to update docs or CI to reflect the changes in this PR yet. We just wanted to get our thoughts down "on paper" for others to review.

config/uaa/enable-automount-service-account-token-for-kind.yml
davewalter: Thanks for the context @jenspinney! I think this will ultimately be decided by the interface that @Syerram wanted to expose to the operator. Personally, I do like having a kind-specific configuration that turns on everything required to deploy.

config/networking/first-party-jwt-istio-for-kind.yml
davewalter: It sounds like we are running into more of these sorts of things, and that we generally don't want to tie them directly to a specific IaaS.

ciriarte: I would favor [feature detection](https://en.wikipedia.org/wiki/Feature_detection_(web_development)) over IaaS/version. This is something I feel we need to discuss in a wider setting if possible. @jamespollard8 @davewalter @nhsieh

loewenstein: I agree that feature detection would be preferable for such cases. As a platform operator I just want the right thing to happen, not find out the right set of config knobs.

ciriarte: e.g. for this particular case:

To determine if your cluster supports third party tokens, look for the TokenRequest API:

```console
$ kubectl get --raw /api/v1 | jq '.resources[] | select(.name | index("serviceaccounts/token"))'
{
    "name": "serviceaccounts/token",
    "singularName": "",
    "namespaced": true,
    "group": "authentication.k8s.io",
    "version": "v1",
    "kind": "TokenRequest",
    "verbs": [
        "create"
    ]
}
```

Syerram:
If I understand the goal of the PR - move away from config-optional to data values interface, which I believe it is accomplishing. However, some of the questions are related to whether they should be exposed or be auto-detected among others. I don't think we should worry about the "ideal" main config just yet. 

I recommend for now we merge with just top-level data values and then we can figure out the right way (whether that is auto-detect k8s, init-script, [insert your soln here]) later. We are still in pre-1.0 world and we have the luxury of moving fast and breaking things.

ci/tasks/install-cf-on-kind/task.sh
```diff
@@ -5,6 +5,14 @@ source cf-for-k8s-ci/ci/helpers/auth-to-gcp.sh
 
 echo "Generating install values..."
 cf-for-k8s/hack/generate-values.sh -d vcap.me -g gcp-service-account.json > cf-install-values/cf-install-values.yml
+cat <<EOT >> cf-install-values/cf-install-values.yml
+add_metrics_server: true
+automount_service_account_token: true
+enable_load_balancer: false
+patch_metrics_server_for_kind: true
```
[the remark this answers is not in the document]
davewalter: I totally agree. I wasn't happy about the name of this flag in particular, but I didn't give it too much thought as I believe @Syerram is planning to revisit the data values interface soon.

config/eirini/enable-automount-service-account-token.yml
```diff
@@ -1,7 +1,8 @@
+#@ load("@ytt:data", "data")
```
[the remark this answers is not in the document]
davewalter: Happy to make these align for now. As I stated, I believe @Syerram is planning to revisit the data values interface soon, so this could all change again next week.

config/networking/remove-resource-requirements.yml
davewalter: I think we came to the same conclusion as a team.

config/metrics/add-metrics-server-components.yml
davewalter: Everything should be up to date now.

config/eirini/enable-automount-service-account-token.yml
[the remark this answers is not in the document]
davewalter: The only one that this doesn't work with is `enable_load_balancer`, which controls the contents of `config/networking/ingressgateway-service-nodeport.yml`.

PR #376, opened 2020-08-19
Minor docs grammar fix (it's -> its)

jenspinney:
## WHAT is this change about?
This is a tiny change to the deploy docs to fix incorrect instances of "it's" where "its" is the correct form.

No

## Acceptance Steps
When I look at the deploy docs, there are no incorrect usages of "it's".


jenspinney:
The failed check seems to be broken. This change shouldn't effect any behavior.

PR #379, opened 2020-08-20
[WIP] add revision info task at beginning of every job

phil9909:
## WHAT is this change about?
It adds a task to the beginning of every job of every pipeline (created by a template) which prints the current git revision

e.g.:
```yaml
  plan:
  - task: pipeline-revision-info
    config:
      platform: linux
      image_resource:
        source:
          repository: busybox
        type: docker-image
      run:
        path: /bin/echo
        args:
        - -e
        - git@github.com:cloudfoundry/cf-for-k8s.git 5c6c847fcdfa25765084d2f659f212c46099cb5f ci/templates/cf-for-k8s-contributions.yml \e[31mdirty\e[0m
```
image

We do this for quite some time now for our internal pipelines and it is helpful from time to time.  

## Limitation

The current implementation is WIP and has a limitation. If you're interested we can find a way to resolve the limitation:

As the result of the templating process is currently check into git, it will always be marked as dirty (as the result contains the hash and the hash is unknown before committing).
We could either stop doing that or run a 2nd `ytt template` only to add this overlay.

No

## Acceptance Steps
run `ci/configure <some pipeline>`

### DIFF
--- a/ci/configure
+++ b/ci/configure
@@ -14,9 +14,16 @@ function main() {
 
 EOF
 
+    REVISION="$(git remote get-url origin) $(git rev-parse HEAD) $(git ls-files --full-name "${PROJECT_DIR}/ci/templates/${PIPELINE}.yml")"
+    if [[ -n $(git status --porcelain) ]] ; then
+      REVISION="$REVISION \e[31mdirty\e[0m"
+    fi
+
     ytt template \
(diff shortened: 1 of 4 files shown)


paulcwarren:
Thanks much @phil9909

PR #381, opened 2020-08-20
docs: add images to routing topology docs

KauzClay:
We moved the docs in but forgot to include the images. This should fix that, sorry!


acosta11:
I had just the same idea when I circled back to look at PR's just earlier! I think I got the fix in #380.

PR #382, opened 2020-08-21
Add validation test for external database

Haegi:
## WHAT is this change about?
With this change, we verify that the configured database is actually used (#168).
We do so by verifying that tables exist in the public schema for `uaa` and `capi`.
This is implemented for in-cluster databases as well as external (rds) databases.

No

@modulo11 
@phil9909

### DIFF
--- a/ci/pipelines/cf-for-k8s-contributions.yml
+++ b/ci/pipelines/cf-for-k8s-contributions.yml
@@ -651,6 +651,16 @@ jobs:
         APP_REGISTRY_USERNAME: ((cf_for_k8s_private_dockerhub.username))
         APP_REGISTRY_PASSWORD: ((cf_for_k8s_private_dockerhub.password))
         USE_EXTERNAL_DB: "true"
+  - task: run-external-db-validation-test
+    file: cf-for-k8s-ci/ci/tasks/run-external-db-validation-test/task.yml
+    params:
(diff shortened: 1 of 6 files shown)


davewalter:
Thanks for the contribution!

davewalter:
Hi @Haegi @modulo11 @phil9909,

Just wanted to follow up on one of the changes from this PR. We ran into problems with the change to make the app registry password accept multi-line strings in the `install-cf-on-gke` task. While we understand that this is desirable for GCP service account JSON keys, it causes our CI to break when we provide a single-line dockerhub password. The failure manifests itself as an inability to push the `cf-default-builder` image to the registry during the install. See [here](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s-contributions/jobs/validate-pr-on-gke/builds/81) for an example failure.

We have reverted the task to expect a single-line password [here](https://github.com/cloudfoundry/cf-for-k8s/commit/c5db1aae78e7c1b4732f29ab45968331f3e7c970) to unblock our CI. We aren't sure if you use these tasks in your own CI so please let us know if this was an intentional change so we can work with you on a solution that works for everyone.

Regards,
Dave

Haegi:
Hi @davewalter, 

Thanks for reverting the change.
We changed the secret on our side from a multi-line string to a single-line string.
Everything should now work for both of us.

Best regards,
Benni

PR #391, opened 2020-08-25
Update metric-proxy to v1.0.0

dtimm:
## WHAT is this change about?
Remove extra TLS layer from CAPI -> metric-proxy communication

No

## Acceptance Steps
Run `cf app <app-name>` and see that CPU/memory usage are reported.

@cloudfoundry/cf-loggregator 
@hev 
@kkburr


davewalter:
Thanks for the contribution!

jamespollard8:
hey @dtimm and @kkburr, it looks like we missed removing the certs from the https://github.com/cloudfoundry/cf-for-k8s/blob/f4dcf7b5c364a648d75e775e47a42c716a7a5fb8/sample-cf-install-values.yml#L98

Would you be willing to open a PR with a fix for that?

Thanks!
relint

heycait:
Sure will do

PR #399, opened 2020-08-27
Allow ingress to cf-api-server from eirini-events

gcapizzi:
## WHAT is this change about?

When an app container crashes, eirini-events will report it to cf-api-server. Create rule to allow this in network policies.

No.

## Acceptance Steps
* Run the following:
  ```
  cf push catnip
  curl <catnip-url>/sigterm/KILL
  cf events catnip
  ```
* See crash events in the output

@cloudfoundry/eirini


davewalter:
Thanks for the contribution!

PR #406, opened 2020-08-31
Make the docs simpler

Syerram:
## WHAT is this change about?
- Remove unnecessary distractions
- Reduce the number of options and instead choose the easiest option (e.g. dockerhub over gcr/acr)
- Add roadmap/milestones at the end

## Acceptance Steps
Running through the steps should install a successful cf-for-k8s

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -82,7 +82,7 @@ To be able to push source-code based apps to your cf-for-k8s installation, you w
    1. Generate certificates for the above domains and paste them in `crt`, `key`, `ca` values
       - **IMPORTANT** Your certificates must include a subject alternative name entry for the internal `*.cf-system.svc.cluster.local` domain in addition to your chosen external domain.
 
-1. To enable Cloud Native buildpacks feature, configure access to an external dockerhub registry in `cf-values.yml` you setup above in the section **Setup a a docker registry**
+1. Configure access to the dockerhub registry in `cf-values.yml` that you setup in the above section [Setup docker registry](#setup-a-docker-registry)
 
       ```yml
 
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -43,8 +43,8 @@ To deploy cf-for-k8s as is, the cluster should:
 To be able to push source-code based apps to your cf-for-k8s installation, you will need to add OCI compliant registry (e.g. hub.docker.com) to the configuration.
 
 [hub.docker.com](https://hub.docker.com/) is pretty easy to get started
-  1. Create an account in [hub.docker.com](https://hub.docker.com/). Note down the user name and password you used during signup.
-  1. Create a repository in your account. Note down the repository name.
+  1. Create an account in [hub.docker.com](https://hub.docker.com/). Note down the **username** and **password** you used during signup. You will use them in deployment steps below.
+  1. Create a repository in your account. Note down the **repository** name.
 
 ## Steps to deploy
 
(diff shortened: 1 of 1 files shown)


docs/deploy.md
```diff
@@ -254,10 +193,10 @@ cf-for-k8s can be configured to [use an external database](platform_operators/ex
 1. Validate the app is reachable over **https**
 
    ```console
-   # for self-signed certs, use -k to allow insecure server connections when using SSL
-   curl -k https://test-node-app.<cf-domain>
+   curl -Lk https://test-node-app.<cf-domain>
    ```
 
+   You should she the following output
```
ctlong, line 199: Small typo: "she" vs "see"

docs/deploy.md
```diff
@@ -119,39 +94,21 @@ cf-for-k8s can be configured to [use an external database](platform_operators/ex
...
 
       ii. Install using `kapp` and pass the above K8s configuration file
       ```console
       kapp deploy -a cf -f ${TMP_DIR}/cf-for-k8s-rendered.yml -y
       ```
-      > cf-for-k8s uses [kapp](https://github.com/k14s/kapp) to manage its lifecycle. `kapp` will first show you a list of resources it plans to install on the cluster and then will attempt to install those resources. `kapp` will not exit untill all resources are installed and their status is running. See all options by running `kapp help`.
 
-   Once you run the command, it should take about 10 minutes depending on your cluster bandwidth, size. `kapp` will provide updates on pending resource creations in the cluster and will wait until all resources are created and running. Here is a sample snippet from `kapp` output:
+   Once you run the command, it should take about ~5-7 minutes depending on your cluster bandwidth, size. `kapp` will provide updates on pending resource creations in the cluster and will wait until all resources are created and running. Here is a sample snippet from `kapp` output:
```
jamespollard8, line 111: I'd leave it at "about 10 minutes"

docs/deploy.md
```diff
@@ -254,10 +193,10 @@ cf-for-k8s can be configured to [use an external database](platform_operators/ex
 1. Validate the app is reachable over **https**
 
    ```console
-   # for self-signed certs, use -k to allow insecure server connections when using SSL
-   curl -k https://test-node-app.<cf-domain>
+   curl -Lk https://test-node-app.<cf-domain>
```
jamespollard8, line 196: I think the `-L` usually isn't important here?

docs/deploy.md
```diff
@@ -213,38 +170,20 @@ cf-for-k8s can be configured to [use an external database](platform_operators/ex
    ```console
    cf push test-node-app -p tests/smoke/assets/test-node-app
    ```
+   
+   You should she the following output from the above command
```
jamespollard8, line 174: typo: `she` > `see`

docs/deploy.md
```diff
@@ -41,29 +38,13 @@ To deploy cf-for-k8s as is, the cluster should:
 - defines a default StorageClass
   - requires [additional config on vSphere](https://vmware.github.io/vsphere-storage-for-kubernetes/documentation/storageclass.html), for example
 
-### Requirements for pushing source-code based apps to Cloud Foundry foundation
+### Setup a docker registry
```
jamespollard8, line 41: FWIW I like having both docker and GCR documented here

docs/deploy.md
jamespollard8: Nope, seems like I'm wrong here. [slack thread](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1598910866049200?thread_ts=1598910828.048900&cid=CH9LF6V1P)

docs/deploy.md
loewenstein: Do we instrument installation to track where the time is spent? ~10 minutes sounds a lot of time (even 5-7 minutes sounds like worth looking into).

jamespollard8: To clarify, we're talking about fresh install time and I think we're feeling rather proud about the relatively fast install time here (especially compared to cf for VMs).

My understanding is that it takes 4+ minutes to go from "fresh cluster" to "ready to `cf push`" largely because there are a few component dependencies (eg UAA requires the database to be up and running, log cache requires UAA, etc).

davewalter: Maybe change it to "about 10 minutes or less, depending on your cluster ..."?

docs/deploy.md
davewalter: The `-L` is important if you specify a URL with an `http` scheme as the ingressgateway responds with a redirect. To the best of my knowledge, it is not required for this example.

PR #410, opened 2020-08-31
update logging to 0.3.0

Benjamintf1:
## WHAT is this change about?
* update cf-k8s-logging to 0.3.0
  * remove tls from log-cache, rely on envoy for tls
  * update source type to look like logs on cf on bosh, include process type and source type
  * remove envoy tailed logs from log output for clarity

yes, we removed tls certs from log-cache!!!! 

## Acceptance Steps
* integration tests still work
* you can see things like "/proc/web" in logs from logs from running apps
* you don't see a bunch of envoy logs on startup of apps
* you don't have to generate tls certs for logging 

@pianohacker @MasslessParticle @JSchuenke


Benjamintf1:
I'm going to fix the conflict

Benjamintf1:
I'm guessing the uptime tests don't work right now

```

Staging app and tracing logs...
FAILED

stderr:
Buildpack "binary_buildpack" must be an existing buildpack configured for use with kpack
```

jamespollard8:
Yep, sorry about that @Benjamintf1 

The upgrade test is broken from our side (related to upgrading the cf-cli). We should have a fix today or tomorrow, we'll rerun the tests for this PR, and will report back.

PR #413, opened 2020-09-01
Remove metric_proxy certs from sample values file

heycait:
[#173597144]

## WHAT is this change about?
Remove metric_proxy from sample values file now that TLS has been removed here: https://github.com/cloudfoundry/cf-for-k8s/pull/391

No


jamespollard8:
FYI I found one more: https://github.com/cloudfoundry/cf-for-k8s/pull/415

PR #418, opened 2020-09-07
Allow Istio loadBalancerSourceRanges to be configured

Infra-Red:
## WHAT is this change about?

Allow to install cf-for-k8s with a loadbalancer that is only accessible to clients with the specific IP addresses. This configuration will be ignored if the cloud-provider does not support the feature.

https://v1-17.docs.kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/

## Does this PR introduce a change to `config/values/00-values.yml`?

Yes.

## Acceptance Steps

Please follow the steps from the document which is part of this PR.

### DIFF
--- a/docs/platform_operators/define-loadbalancer-firewalls.md
+++ b/docs/platform_operators/define-loadbalancer-firewalls.md
@@ -1,0 +1,51 @@
+# Configure Cloud Provider's Firewalls for Istio Loadbalancer
+
+## Objective
+
+At end of this setup, you will be able to install cf-for-k8s with a loadbalancer that is only accessible to clients with the specific IP addresses.
+
+## Prerequisites
+
+- Cluster should support `LoadBalancer` services.
+
+## Steps to setup cloud provider firewalls
+
+The following instructions assume you have created `cf-install-values.yml`.
+
+1. Add `istio_source_ranges` key and the IP ranges that are allowed to access the load balancer to your `cf-install-values.yml`
+
+    > Note that the append annotation must be applied to each item you want to insert.
+
+    ```yaml
+    istio_source_ranges:
+    #@overlay/append
+    - "130.211.204.1/32"
+    #@overlay/append
+    - "130.211.204.2/32"
+    ```
+
+1. Follow the instructions from [deploy doc](../deploy.md) to generate the final deploy yml using `ytt` and then `kapp` deploy cf-for-k8s to your cluster.
+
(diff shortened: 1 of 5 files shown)


docs/platform_operators/define-loadbalancer-firewalls.md
```diff
@@ -0,0 +1,51 @@
...
+
+    ```yaml
+    istio_source_ranges:
+    #@overlay/append
+    - "130.211.204.1/32"
+    #@overlay/append
+    - "130.211.204.2/32"
+    ```
+
+1. Follow the instructions from deploy doc to generate the final deploy yml using `ytt` and then `kapp` deploy cf-for-k8s to your cluster.
```
jamespollard8, line 27: It'd be nice to provide a hyperlink to the deploy doc here

jamespollard8 (approved):
Looks great to us (cc @ericpromislow). We'll leave it to our PM to do acceptance and then merge this PR.

docs/platform_operators/define-loadbalancer-firewalls.md
Infra-Red: Thank you for your feedback! I've updated PR accordingly.

paulcwarren:
Hi @Infra-Red, many thanks for your contribution.  Very much appreciated.

Unfortunately, on this occasion, we will not be able to accept your PR.  Moving forwards our intention is actually to remove the end-user facing `istio_static_ip` configuration property altogether and as a result we are unable to accept a PR that would add a second istio-related configuration property.

Very briefly, we aim to support native K8s API so that we are IaaS agnostic and can support a wide range of K8s distributions.  K8s doesn't yet offer a load balancer API and configurations pertaining to load balancer therefore require IaaS specific configuration.  In this case, we have found that `istio_static_ip` is somewhat GKE specific and unworkable on AKS for example.  By way of mitigation we do plan to document how to configure cf-for-k8s for use with common IaaS load balancers.  

Secondarily, this would expose the internals of our networking component making it part of the cf-for-k8s configuration API that we would then have to support moving forwards and that is also not something that we are prepared to do at this time.
You can find some more details outlined in this [Github Issue](https://github.com/cloudfoundry/cf-for-k8s/issues/420).

We therefore recommend that you re-purpose this code as an optional overlay.  

I apologize for any inconvenience that this may have caused.  We are trying to get better at communicating product roadmap more effectively to avoid situations such as these.

PR #425, opened 2020-09-08
remove capi eirini cert config

cwlbraa:
this allows us to remove cert configuration and bogus defaults from capi-k8s-release default values files.

## WHAT is this change about?
This removes a few unnecessary properties from capi.yml that used to configure TLS communication between CAPI and eirini.

No.

## Acceptance Steps
It should deploy without a hitch. After this is merged, please poke us to merge cloudfoundry/capi-k8s-release#68 .

### DIFF
--- a/config/capi/capi.yml
+++ b/config/capi/capi.yml
@@ -43,16 +43,6 @@ ccdb:
   database: #@ data.values.capi.database.name
   ca_cert: #@ data.values.capi.database.ca_cert
 
-eirini:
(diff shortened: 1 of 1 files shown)


jamespollard8:
Hey @cwlbraa, looks like we hit a concourse glitch here. Could you please do us a favor and just open a new PR for this?

Sorry for the inconvenience - the other toggles we tried don't actually help concourse detect this PR/branch/commit.

PR #427, opened 2020-09-09
Define scale properties in main config

ericpromislow:
## WHAT is this change about?

https://www.pivotaltracker.com/story/show/174586782

## Does this PR introduce a change to `config/values.yml`?
Yes (`config/values/00-values.yml`)

## Acceptance Steps

This change is mainly to define changes to the interface. It's sufficient to deploy this branch and run smoke tests successfully.

## Tag your pair, your PM, and/or team
Eric Promislow & @davewalter 

## Things to remember
[x] Make sure this PR is based off the `develop` branch
[x] Include any links to related PRs, issues, stories, slack discussions, etc... that will help establish context.
[x] No: Is there anything else of note that the reviewers should know about this change?
[x] None: - Our CI uses concourse task files from HEAD of the `develop` branch. If your PR includes/requires CI changes, please ping the RelInt interrupt and they'll help apply the CI changes for you.
[x] Read: - Checkout the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/community/CONTRIBUTING.md)

PR #429, opened 2020-09-10
Move Istio config and build scripts to cf-for-k8s

ndhanushkodi:
* They used to be in cf-k8s-networking

[#174370911](https://www.pivotaltracker.com/story/show/174370911)

Co-authored-by: Mikael Manukyan <mmanukyan@vmware.com>

## WHAT is this change about?
This change moves the Istio related configuration and build scripts to cf-for-k8s. There was a pain point on the networking team around consistently having cf-k8s-networking develop branch be incompatible with cf-for-k8s, and that made it difficult to work on multiple stories at once. Istio is a large source of those incompatibilities, so moving Istio to cf-for-k8s and making the changes directly here could help alleviate that.

No

## Acceptance Steps
When I deploy cf-for-k8s with this PR,
the same set of smoke tests should continue to pass.
(No functional changes)

pair: @mike1808 

cc @keshav-pivotal @shalako 

## Notes
- We're looking for feedback from the Release Integration team on structure/readability on this PR while it's in draft form

~~This change is currently based on the `ingressgateway-fluent-bit-sidecars` branch, since we would like that change merged first. It will be rebased onto develop before being marked 'ready for review'.~~ EDIT: rebased on top of develop


config/values/30-istio-values.yml
```diff
@@ -0,0 +1,11 @@
+#@data/values
+#@overlay/match-child-defaults missing_ok=True
+---
+istio:
+  #! build_version is updated by build/istio/build.sh, do not edit it manually
+  build_version: "1.6.4"
```
Birdrock, line 6: There is a concern that this exposes the apparent istio version to the operator - they may think they can change it via this value. Perhaps it could be insulated from the user in your library eval value replacement step or replaced in another way?

config/networking/istio/README.md
```diff
@@ -0,0 +1,9 @@
+## Istio configuration
```
Birdrock, line 1: We feel that this should probably be a separate directory inside `config`, instead of a sub-directory of `networking`.

config/values/30-istio-values.yml
```diff
@@ -0,0 +1,11 @@
...
+  upgrade_sidecars:
+    image: gcr.io/cf-networking-images/cf-k8s-networking/upgrade-sidecars@sha256:8cef1082ec0eb1d2e2f78c9b668065ce106de775cec929c745ca0e7f20a4c8fd
+
+  istiod:
+    replicas: 1
```
Birdrock, line 11: Please see PR #427 for our proposed scaling interface.

Birdrock (commented):
We took a look at it and made some suggestions. Would you like to chat or have a x-team pair session?

config/values/30-istio-values.yml
XanderStrike: Is there a place where we can put "values used by multiple overlays but not editable by users?" We need this in a few places (changing it and rerunning our scripts is actually how we upgrade istio), but we agree we don't want users to think they can edit it.

The same goes for the image shas below, and the istiod replicas. These are internal values that users shouldn't be editing but need to be in one place to DRY up our overlays.

config/values/30-istio-values.yml
```diff
@@ -0,0 +1,11 @@
+#@data/values
+#@overlay/match-child-defaults missing_ok=True
+---
+istio:
+  #! build_version is updated by build/istio/build.sh, do not edit it manually
+  build_version: "1.6.4"
+  upgrade_sidecars:
+    image: gcr.io/cf-networking-images/cf-k8s-networking/upgrade-sidecars@sha256:8cef1082ec0eb1d2e2f78c9b668065ce106de775cec929c745ca0e7f20a4c8fd
```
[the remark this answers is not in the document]
XanderStrike: See comment here: https://github.com/cloudfoundry/cf-for-k8s/pull/429#discussion_r486690628

config/values/30-istio-values.yml
XanderStrike: See comment here: https://github.com/cloudfoundry/cf-for-k8s/pull/429#discussion_r486690628

build/istio/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -eu
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+${SCRIPT_DIR}/generate.sh "$@" | kbld -f - > "${SCRIPT_DIR}/../../config/networking/istio/istio-generated/xxx-generated-istio.yaml"
```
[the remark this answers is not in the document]
ndhanushkodi: When developing sometimes we'd want to run just generate.sh and see the result, but not necessarily save it to the generated istio config (ie for tweaking a property and seeing if the ytt commands still run)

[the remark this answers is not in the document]
XanderStrike: Running `./generate.sh | less` is a crucial part of my workflow when iterating on istio config so I'd vote for keeping them separate.

build/istio/overlays/add-fluent-bit-sidecar-to-ingressgateway.yaml
```diff
@@ -0,0 +1,49 @@
...
+spec:
+  template:
+    spec:
+      containers:
+      #@overlay/append
+      - name: fluent-bit
+        image: #@ data.values.fluentbit.image
+        resources:
+          limits:
+            memory: 100Mi
```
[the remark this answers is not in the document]
ndhanushkodi: yup, just added it

config/networking/istio/README.md
ndhanushkodi: I think that makes sense, we made this change just now :)

build/istio/build.sh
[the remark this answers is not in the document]
Birdrock: That makes sense.

config/values/30-istio-values.yml
XanderStrike: We've now eliminated the need for these values:

* `build_version` -- we defined a function in the file it's used to return the value, and updated our build script to update its value
* `istiod.replicas` -- hardcoded while we wait for https://github.com/cloudfoundry/cf-for-k8s/pull/427
* `upgrade_sidecars.image` -- hardcoded with a future chore for our image pipeline to update it in place (this one doesn't change much)

davewalter (approved):
LGTM!

PR #430, opened 2020-09-10
Use Image Registry for Package Uploads and Secure Internal CF API Endpoints

tcdowney:
## What is this change about?
This change does the following:
* Configures Cloud Controller to use the same OCI image registry that kpack uploads droplets to for package uploads (https://github.com/cloudfoundry/cf-for-k8s/issues/409)
* Moves the `/internal` CF API endpoints to a separate port
* Adds additional `NetworkPolicy` and Istio `AuthorizationPolicy` resources to further protect these internal endpoints
* Temporarily updated `vendir.yml` to point to a branch of capi-k8s-release.

After this is merged, we will merge the capi-k8s-release branch (`secure-cf-api-internal-endpoints-174455375`) and y'all can re-enable the auto-bumping of capi-k8s-release in your CI.

## Does this PR introduce a change to `config/values.yml`?
Nope

## Acceptance Steps
Run smoke tests. `cf push` should still work. 🙂

## Tag your pair, your PM, and/or team
@jspawar @cloudfoundry/cf-capi 

## Related links
- https://github.com/cloudfoundry/cf-for-k8s/issues/409
- https://www.pivotaltracker.com/story/show/174455375
- https://www.pivotaltracker.com/story/show/173343000

PR #434, opened 2020-09-14
Validate s3 blobstore

modulo11:
Depends on #478 

## WHAT is this change about?
With this change we verify that an external (Amazon S3) blobstore is actually used. The test is ran for the develop branch, as well as for each PR.

No.

## Acceptance Steps
Set the pipeline and try it out.

@kramerul 
@loewenstein

### DIFF
--- a/ci/pipelines/cf-for-k8s-main.yml
+++ b/ci/pipelines/cf-for-k8s-main.yml
@@ -735,75 +735,106 @@ jobs:
   serial: true
   public: true
   plan:
-    - do:
-      - get: cf-for-k8s-ci
-      - get: cf-for-k8s-s3-terraform-templates
-
-      - task: create-s3-tf-vars-file
-        file: cf-for-k8s-ci/ci/tasks/create-s3-tf-vars-file/task.yml
-        params:
-          AWS_ACCESS_KEY_ID: ((ci_k8s_aws_access_key_id))
-          AWS_SECRET_ACCESS_KEY: ((ci_k8s_aws_secret_access_key))
-          AWS_REGION: ((ci_k8s_aws_region))
-
-      - put: terraform-s3
-        params:
-          terraform_source: cf-for-k8s-s3-terraform-templates/deploy/s3/terraform
-          delete_on_failure: true
-          var_files: [tf-vars/input.tfvars]
-
-      - task: install-s3-buckets
-        params:
-          AWS_ACCESS_KEY_ID: ((ci_k8s_aws_access_key_id))
-          AWS_SECRET_ACCESS_KEY: ((ci_k8s_aws_secret_access_key))
-          AWS_REGION: ((ci_k8s_aws_region))
-        file: cf-for-k8s-ci/ci/tasks/install-s3-buckets/task.yml
-
-    - do:
+    - in_parallel:
       - get: cf-for-k8s-develop
         passed:
           - test-vendir-sync-on-cf-for-k8s
           - run-unit-tests
         trigger: true
-      - get: terraform-s3
+      - get: cf-for-k8s-ci
+      - get: cf-for-k8s-s3-terraform-templates
 
-      - put: ready-pool
-        params:
-          acquire: true
+    - put: ready-pool
+      params:
+        acquire: true
 
-      - task: install-cf
-        input_mapping:
-          pool-lock: ready-pool
-          cf-for-k8s: cf-for-k8s-develop
+    - task: create-s3-tf-vars-file
+      file: cf-for-k8s-ci/ci/tasks/create-s3-tf-vars-file/task.yml
+      params:
+        AWS_ACCESS_KEY_ID: ((ci_k8s_aws_access_key_id))
+        AWS_SECRET_ACCESS_KEY: ((ci_k8s_aws_secret_access_key))
+        AWS_REGION: ((ci_k8s_aws_region))
+
+    - put: terraform-s3
+      params:
+        terraform_source: cf-for-k8s-s3-terraform-templates/deploy/s3/terraform
+        delete_on_failure: true
+        var_files: [tf-vars-s3/input.tfvars]
+
+    - task: install-s3-buckets
+      params:
+        AWS_ACCESS_KEY_ID: ((ci_k8s_aws_access_key_id))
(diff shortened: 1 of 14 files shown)


modulo11:
The errors which are detected in the pipeline for this PR can not be caused by our change, because our change needs to be applied manually by using `fly`.

ci/tasks/install-s3-buckets/task.yml
```diff
@@ -0,0 +1,25 @@
...
+params:
+  AWS_ACCESS_KEY_ID: 
+  AWS_SECRET_ACCESS_KEY: 
+  AWS_REGION:
+
+outputs:
+  - name: blobstore-metadata
+
+run:
+  path: cf-for-k8s-ci/ci/tasks/install-s3-buckets/task.sh
```
davewalter, line 25: It feels like this task should be called something like `create-s3-bucket-data-values-file` to better reflect its purpose.

ci/pipelines/cf-for-k8s-main.yml
```diff
@@ -655,6 +674,111 @@ jobs:
...
+        GCP_PROJECT_ZONE: ((ci_k8s_gcp_project_zone))
+        APP_REGISTRY_HOSTNAME: https://index.docker.io/v1/
+        APP_REGISTRY_REPOSITORY_PREFIX: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_USERNAME: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_PASSWORD: ((cf_for_k8s_private_dockerhub.password))
+        USE_EXTERNAL_BLOBSTORE: true
+        DOMAIN: k8s-dev.relint.rocks
+      file: cf-for-k8s-ci/ci/tasks/install-cf-on-gke/task.yml
+
+    - task: run-external-blobstore-validation-test
```
davewalter, line 730: Could you please run smoke-tests in parallel to this task?

ci/templates/cf-for-k8s-contributions.yml
```diff
@@ -663,6 +682,131 @@ jobs:
...
+        GCP_SERVICE_ACCOUNT_JSON: ((ci_k8s_gcp_service_account_json))
+        DOMAIN: k8s-dev.relint.rocks
+        USE_EXTERNAL_APP_REGISTRY: true
+        APP_REGISTRY_HOSTNAME: https://index.docker.io/v1/
+        APP_REGISTRY_REPOSITORY_PREFIX: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_USERNAME: ((cf_for_k8s_private_dockerhub.username))
+        APP_REGISTRY_PASSWORD: ((cf_for_k8s_private_dockerhub.password))
+        USE_EXTERNAL_BLOBSTORE: true
+
+    - task: run-external-blobstore-validation-test
```
davewalter, line 751: Could you please run smoke-tests in parallel to this task?

davewalter (commented):
Hi @modulo11,

Thanks for contributing these changes to CI. I left a few comments inline.

Regards,
Dave

modulo11:
All comments should be addressed. Is there anything left from our side?

davewalter:
Hi @modulo11,

One last thing I think we need to address revolves around the new `run-external-blobstore-validation-test` test. We recently merged PR #430, which reduces CAPI's use of the blobstore to app bits larger than 64k. This meant that this test was failing in the `cf-for-k8s-contributions/validate-pr-on-gke-external-blobstore` and `cf-for-k8s-main/validate-cf-for-k8s-gke-external-blobstore` jobs as the small node app we use in smoke tests was not large enough to cause the blobstore to be used. To unblock ourselves, we removed the test from these jobs and replaced it with simply running smoke-tests to confirm that the platform is operational.

I discussed this problem with the team and the consensus was that we could probably replace it with a [unit test](https://github.com/cloudfoundry/cf-for-k8s/tree/develop/tests/ytt) that confirms that we are providing the correct blobstore configuration to CAPI. Before we prioritise the work to implement the required Gomega matchers in our [library](https://github.com/cloudfoundry/yttk8smatchers), we wanted to check in with you to make sure that this is an acceptable solution for your team?

Regards,
Dave

loewenstein:
Hi @davewalter, at least for the external database tests we provided one particular motivation to test with in cluster and out of cluster external database was testing networking (i.e. effects of NetworkPolicy and Envoy sidecars). 

Not sure if the same reasoning should lead us to do the same for blobstore. NetworkPolicy is probably covered with the external database tests. Not sure if Envoy could lead to specific problems with blobstore that we'd like to see early on.

davewalter:
Hi @loewenstein, That makes total sense. I think the challenge here is that, with the switch to using the registry for the package bits, the blobstore isn't actually being exercised by the current CI tests. I chatted with the CAPI team and they pointed me to a [Tracker story](https://www.pivotaltracker.com/story/show/174741035) about making the blobstore optional that they expect to get to in a month. Given this, I am not sure if it is worth putting a lot of effort into making this test work. Let me know what you think.

loewenstein:
Hi @davewalter, I tend to agree, that blobstore as a dependency to cf is in its final stage before extinction. It is still going to be shipped with cf-for-k8s 1.0 and given the timelines it might be shipped as required dependency. Or do you consider the work as release blocking?

If we would just increase the app size in CI tests (either in general or by adding a simple test that pushes an app big enough to trigger resource matching) we should be fine, shouldn't we?

davewalter:
Hi @loewenstein,

> It is still going to be shipped with cf-for-k8s 1.0 and given the timelines it might be shipped as required dependency. Or do you consider the work as release blocking?

It sounds like this is not going to land before 1.0. CAPI gave me a timeframe of "October" to work on making the blobstore optional, which is right around our release time for CF Summit.

> If we would just increase the app size in CI tests (either in general or by adding a simple test that pushes an app big enough to trigger resource matching) we should be fine, shouldn't we?

Sounds fine. Is that something you could PR in?

modulo11:
As discussed in #478, this PR is now up2date and can be used in favor.

davewalter (approved):
Looks good to @Birdrock and myself. Thanks!

PR #435, opened 2020-09-14
remove HPA and fix istiod replicas to 1

ndhanushkodi:
Co-authored-by: Kauana dos Santos <kdossantos@vmware.com>
Co-authored-by: Alex Standke <astandke@vmware.com>
Co-authored-by: Clay Kauzlaric <ckauzlaric@vmware.com>
Co-authored-by: Nitya Dhanushkodi <ndhanushkodi@pivotal.io>

## WHAT is this change about?
- the ingressgateway one does not function, because we deploy a
daemonset
- we know from previous research that hpas do not help overloaded
pilot/istiods
- defaults to 1 istiod for proof-of-concept/local deployments

[#174764287](https://www.pivotaltracker.com/story/show/174764287)

No

## Acceptance Steps
When I deploy cf-for-k8s with this PR
I can see that istiod has 1 replica with `kubectl get pods -n istio-system`

@kauana @KauzClay @XanderStrike

### DIFF
--- a/config/istio/remove-hpas-and-scale-istiod.yml
+++ b/config/istio/remove-hpas-and-scale-istiod.yml
@@ -1,0 +1,13 @@
+#@ load("@ytt:overlay", "overlay")
+
+#@ load("@ytt:data", "data")
+
+#@overlay/match by=overlay.subset({"kind":"HorizontalPodAutoscaler"}), expects="1+"
+#@overlay/remove
+---
+
+#@overlay/match by=overlay.subset({"kind": "Deployment", "metadata":{"name":"istiod"}}),expects=1
+---
+spec:
+  #@overlay/match missing_ok=True
+  replicas: 1


config/values/00-values.yml
```diff
@@ -11,6 +11,10 @@ cf_admin_password:
 #! reserved static ip for istio LoadBalancer
 istio_static_ip: ""
 
+#! for highly available deployments, increase to 2
+istiod:
```
mike1808, line 15: should we put all istio configs under one field called `istio`?

jamespollard8: Yeah, that sounds good to us. For `istio_static_ip`, we can keep that one separate and we'll plan on renaming that key at a later data ([story](https://www.pivotaltracker.com/story/show/174852214))

jamespollard8 (changes requested):
We're thinking that we'll solve Production scalability of the platform as one single piece/solution. So that means for this PR it'd be great if you want to set the default number of replicas to 1 but we'd prefer if you could remove the documentation / in-line comments referring to production deployments and scaling up.

config/istio/remove-hpas-and-scale-istiod.yml
```diff
@@ -0,0 +1,13 @@
...
+
+#@overlay/match by=overlay.subset({"kind":"HorizontalPodAutoscaler"}), expects="1+"
+#@overlay/remove
+---
+
+#@overlay/match by=overlay.subset({"kind": "Deployment", "metadata":{"name":"istiod"}}),expects=1
+---
+spec:
+  #@overlay/match missing_ok=True
+  replicas: #@ data.values.istiod.replicas
```
jamespollard8, line 13: hoping to see this hard-coded to 1 for now

config/values/00-values.yml
jamespollard8: But given our primary review comment here, we're requesting that you not include the configurability as part of this PR - please remove the added data value

jamespollard8:
For context, here's our [Scaling Interface proposal doc](https://docs.google.com/document/d/1VwOdsKXUhC47Xw7Cutj7Fx-NuYTSV-diZB8Qkt61xfw/edit); we're currently leaning towards option #2

mike1808:
Hey @jamespollard8 . We've updated the PR.

jamespollard8:
Great, thanks. We'll fix our CI and then re-run the tests here.

jamespollard8 (approved):
These file changes look good to me - just waiting on tests to pass now. 👍

PR #436, opened 2020-09-14
bump capi-k8s-release to use OCI registry for packages

tcdowney:
Trying out just the image registry changes from https://github.com/cloudfoundry/cf-for-k8s/pull/430 to see if that passes the KinD cluster tests.


tcdowney:
Nope, same issues.

Issue #431, opened 2020-09-10
cf_system_namespace, cf_workloads_namespace and cf_staging_namespace configuration properties are broken


paulcwarren:
## Describe the bug
As it stands today `cf_system_namespace`, `cf_workloads_namespace` and `cf_staging_namespace` are [exposed as data values](https://github.com/cloudfoundry/cf-for-k8s/blob/ac0babc9bf2d3b7e1f7f23091bd13e1940efc196/config/values/10-system-values.yml#L4) and are therefore configurable by Rachel, the operator. 

## To Reproduce*
However, if we try to configure any of these and deploy: 

cf-values.yml
```
#@data/values
---
system_namespace: foo
...
```

`kapp deploy -a cf -f <(ytt -f config -f /tmp/${K8S_ENV}-cf-values.yml)`

We run into errors:
```
kapp: Error: Applying create configmap/eirini-ver-1 (v1) namespace: cf-system:
  Creating resource configmap/eirini-ver-1 (v1) namespace: cf-system: namespaces "cf-system" not found (reason: NotFound)
```
Because these namespaces are actually hardcoded in a number of places.

## Expected behavior

We should remove this unfinished capability until we understand what use cases it is serving.

# Dev Notes
- If possible, it may make sense to refactor our ytt overlays to use functions to provide these values rather than have them hardcoded in lots of places

PR #438, opened 2020-09-15
Replace namespace data values with internal functions

davewalter:
## WHAT is this change about?
Replace namespace data values with internal functions

See #431

Yes

## Acceptance Steps
Deploy cf-for-k8s and run smoke-tests

### DIFF
--- a/config/capi/capi.yml
+++ b/config/capi/capi.yml
@@ -4,6 +4,7 @@
 #@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:template", "template")
 #@ load("@ytt:yaml", "yaml")
+#@ load("/namespaces.star", "system_namespace", "workloads_namespace", "workloads_staging_namespace")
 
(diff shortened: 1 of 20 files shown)


davewalter:
> Hello friend, it looks like your pull request has failed one or more of our checks. Please take a look! 👀

That is unfortunate. Looks like we need to update our test framework to support top-level library functions when selectively rendering individual components for testing.

paulcwarren:
> > Hello friend, it looks like your pull request has failed one or more of our checks. Please take a look! 👀
> 
> That is unfortunate. Looks like we need to update our test framework to support top-level library functions when selectively rendering individual components for testing.

That's a pain!  I really don't want to go back to one bug directory only.

ericpromislow:
It makes sense that we needed to explicitly include namespaces.star in the test file.  We could add a `common` directory to `config` and add `config/common` to the list of templates to load in each test.

PR #439, opened 2020-09-15
Update the sample-cf-install-values.yml blobstore key

davewalter:
## WHAT is this change about?
The blobstore key in the sample-cf-install-values.yml file was out of sync with the corresponding data values key in config/values/00-values.yml

No

## Acceptance Steps
Use the sample-cf-install-values.yml file to hand-craft a configuration file to deploy with an external blobstore


ericpromislow:
[finishes #174812849](https://www.pivotaltracker.com/story/show/174812849)

PR #461, opened 2020-09-21
Make cc log level configurable

matt-royal:
## WHAT is this change about?
Allow the operator to change the log level of the Cloud Controller components (i.e. cf-api-server, cf-api-worker, cf-api-clock)

Yes

## Acceptance Steps
This command lists all of the log levels present in CC containers:
```
 kapp -a cf logs --pod-name 'cf-api-%' | grep log_level | cut -d '|' -f 2- | jq -r .log_level | sort -u
```

Before deploying this change, this should only list `info`.

To verify this change, add the `capi.log_level` property to your `values.yml` with a value of `"debug"`. After deploying with that values file, the above `kapp` command should list both `info` and `debug`.

@piyalibanerjee @pivotal-mikegresham 

## Things to remember
Relevant tracker story: https://www.pivotaltracker.com/story/show/172847921

### DIFF
--- a/config/capi/capi.yml
+++ b/config/capi/capi.yml
@@ -35,6 +35,9 @@ blobstore:
   buildpack_directory_key: #@ data.values.blobstore.buildpack_directory_key
   aws_signature_version: #@ data.values.blobstore.aws_signature_version
 
+cc:
+  log_level: #@ data.values.capi.log_level
+
 ccdb:
   adapter: #@ data.values.capi.database.adapter
   host: #@ capi_host()
--- a/config/values/00-values.yml
+++ b/config/values/00-values.yml
@@ -46,6 +46,7 @@ capi:
     name: cloud_controller
     #! Plain text ca certificate if tls should be used
     ca_cert: ""
+  log_level: info
 
 uaa:
   database:


cppforlife:
instead of exposing this data value in cf-for-k8s you can just use ytt's ability to configure pulled in libraries directly. add following snippet to your user data values file:

```yaml
#@data/values
#@library/ref "capi-k8s-release"
---
cc:
  log_level: info
```

above tells ytt to configure cc.log_level data value for capi-k8s-release (https://github.com/cloudfoundry/cf-for-k8s/pull/461/files#diff-de336792dc37e21dd352c22f6fb33480R14-R15)

matt-royal:
@cppforlife Thanks for teaching us that trick! That will be very useful for our dev templates. Our intention for this PR is to make it easier for customers to get debug logging for CC. Given that, we think that the explicit property is a simpler/clearer UX for them than asking them to type out a custom overlay. We could also commit the overlay file to cf-for-k8s as an optional template, but that doesn't seem much simpler to us.

matt-royal:
@jamespollard8 @davewalter It looks like the PR test failure is unrelated to our change. Is that right? Please let us know what we can do to help get this green.

acosta11:
> It looks like the PR test failure is unrelated to our change. Is that right?
Correct. This looks like a Dockerhub throttling issue preventing a docker push. Taking a look now to get that test re-run.

paulcwarren:
@matt-royal Did a whole bunch of extra stuff land in this PR?   As I look through the code a lot of it seems unrelated to configuring log level.

acosta11:
> Did a whole bunch of extra stuff land in this PR? As I look through the code a lot of it seems unrelated to configuring log level.

I think we need to rebase on develop when we get the latest capi image update merged for the kpack bump. This looks like the current diff between the kpack branch and the current capi ci-passed branch.

jamespollard8:
@matt-royal Could you please rebase off latest develop?

matt-royal:
@jamespollard8 I rebased the branch against latest develop

@paulcwarren I had bumped the version of capi-k8s-release in the vendir.yml to take in some changes there that were required for this PR. That's not necessary now that your CI has bumped capi-k8s-release

jamespollard8:
> @cppforlife Thanks for teaching us that trick! That will be very useful for our dev templates. Our intention for this PR is to make it easier for customers to get debug logging for CC. Given that, we think that the explicit property is a simpler/clearer UX for them than asking them to type out a custom overlay. We could also commit the overlay file to cf-for-k8s as an optional template, but that doesn't seem much simpler to us.

@paulcwarren Can you please help us make the call on this interface decision? I think the primary question is: "is this worth adding to our values file?"

davewalter (commented):
From a technical perspective, does this setting apply to all CF API components, or just CC? If the latter, perhaps a more focused setting (such as `cc_log_level`) would be appropriate?

Birdrock:
I think setting log levels is a concern that is relevant from seasoned users to people kicking the tires. For this reason, I'd advocate pulling it up from an overlay to a highly visible data value as in this PR. I'd welcome discussion about detriments.

paulcwarren:
Personally, I really like the direct configuration approach; i.e.

```
#@data/values
#@library/ref "@capi-k8s-release"
---
cc:
  log_level: debug
```

as I see the wiring of this underlying property into a cf-for-k8s property as somewhat redundant boilerplate.

The other point here is that this wouldn't need to be a separate overlay *file*.  It can be an additional overlay in the cf values file so really, in terms of effort, it is about typing a few extra lines of code. 

That said I think the problem with the direct configuration approach at the moment is discoverability.  Component's properties are not documented anywhere (that I am aware of).  We also shouldn't be documenting them manually either.  We should really be generating those docs from the component's data values schema.  Plus I doubt our user are really even that cognizant of the fact that different component exists and are separately configurable (any notion of components being very much a design-time construct atm).  When ytt schema lands and is picked up by the component team's and we have an epic of work to generate holistic API documentation from it and publish it somewhere (all things I would like to see happened but not currently planned for, much less agreed to) I think this PR to expose the property is ok.  Practically, speaking I also think there is a subset of properties that we would always wire into cf-for-k8s in order to create a consistent UX and this most likely fits that category.  

So much as I might like to use the direct config approach I don't think we are at a place to implement that quite yet.

matt-royal:
> So much as I might like to use the direct config approach I don't think we are at a place to implement that quite yet.

@paulcwarren Does that mean you're alright with the approach taken in this PR for now?

---

Everyone: Now that the checks are passing, are there any objections to merging this as is?

jamespollard8 (approved):
LGTM - thanks @matt-royal for your patience here

jamespollard8:
@matt-royal How do I discover the log_level options?

Do you think defaulting cf-for-k8s to `debug` for now would be reasonable, given that we're targeting tire-kickers for our 1.0 release?

PR #464, opened 2020-09-21
Expect ingressgateway to be either a Deployment or DaemonSet

kauana:
- Deployments should be an option too if user wants to scale their cluster up and having an istio-ingressgateway per node is too much.

[#171678788](https://www.pivotaltracker.com/story/show/171678788)

Co-authored-by: Leah Hanson <lhanson@pivotal.io>

## Acceptance Steps
**When** I configure the istio-ingressgateway to be a Deployment
**Then** I expect the deployment to be successful

@astrieanna

### DIFF
--- a/build/istio/istioctl-values.yaml
+++ b/build/istio/istioctl-values.yaml
@@ -15,7 +15,7 @@ spec:
             maxReplicas: 1
           service:
             externalTrafficPolicy: Local
-            ports: # TODO: remove when https://github.com/istio/istio/issues/24432 fixed
+            ports: #! TODO: remove when https://github.com/istio/istio/issues/24432 fixed
             - name: status-port
               port: 15021
               targetPort: 15021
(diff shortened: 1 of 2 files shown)


Birdrock:
@kauana How would one change it to a deployment? If it is via overlay, maybe the overlay example to do so would also include this overlay bit to enable it for deployments.

kauana:
Hey @Birdrock,

Currently, we don't have a way to allow platform operators to deploy their ingress gateway as either a DaemonSet or a Deployment. 

We've added the ability to match by either a DaemonSet or a Deployment because on default deployments we use a DaemonSet, but for our scaling cluster (100 nodes) a DaemonSet is overkill so we use ~10 ingress gateways instead.

I do, however, believe that we should make this configurable to the operators and we should have a scaling recommendation doc in case operators choose to go with a Deployment. I am [currently talking](https://cloudfoundry.slack.com/archives/CFX13JK7B/p1600966822011100) to our PMs and we should have stories accomplish that soon.

Thanks

jamespollard8 (approved):
LGTM

PR #469, opened 2020-09-23
Bump capi-k8s-release to fix docker lifecycle apps

jspawar:
## WHAT is this change about?
Accidentally regressed functionality in our first attempt to restrict access to CC's internal endpoints which we introduced in an earlier PR. This regression prevents docker lifecycle apps from working (i.e. `cf push dockerapp -o <some_docker_image>`)

It is worth noting that this change updates vendir to revert capi-k8s-release to pointing at our `ci-passed` branch which we originally intended to be consumed before all of our various feature PRs we had recently (using image registry for packages, restricting internal endpoint access, and upgrading to latest kpack). The branch in its current state for this PR includes all the changes from those other PRs still of course as well as the fix for docker lifecycle apps.

No

## Acceptance Steps
1. Confirm that `cf push buildpackapp` still works and the resulting app is routable
1. Confirm that `cf push dockerapp -o <some_docker_image>` still works and the resulting app is routable

@cloudfoundry/cf-capi

### DIFF
--- a/config/capi/_ytt_lib/capi-k8s-release/templates/nginx-configmap.yml
+++ b/config/capi/_ytt_lib/capi-k8s-release/templates/nginx-configmap.yml
@@ -72,7 +72,7 @@ data:
       server {
           listen 9023;
 
-          location /internal/v4/ {
+          location /internal/ {
             access_log  /cloud_controller_ng/nginx-access.log;
             proxy_buffering             off;
             proxy_set_header            Host $host;
--- a/config/capi/_ytt_lib/capi-k8s-release/values/images.yml
+++ b/config/capi/_ytt_lib/capi-k8s-release/values/images.yml
@@ -1,9 +1,9 @@
 #@data/values
 ---
 images:
-  ccng: gcr.io/cf-capi-arya/dev-ccng@sha256:fda89ac251d0b0aa6185b4b46d42a11113fafff4a6aa2fbc99c4227fc14f35b9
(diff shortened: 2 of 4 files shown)


jspawar:
Hmm, it looks like two of those three failures were timeouts during staging and the third was some unknown failure to send traffic to the pushed app. Haven't observed that on our end though when we run smoke tests though.

Anything we should try to see if we can reproduce?

ericpromislow:
Hi @jspawar 

We have had some problems with the pipeline lately, mostly due to problems deleting the postgres pod on a newer version of k8s, but here's the rundown on your PR failures:

* concourse-ci/external-blobstore: app failed to push in 5 minutes.  Usually it's a kpack error, or upstream from kpack

* concourse-ci/upgrade-check-uptime  - another staging error, and the db-delete failure

ericpromislow:
Could you resolve the conflicts?

The fix passes the ACs above, and we're ignoring the failed pipeline builds for now because we're seeing them elsewhere.

jspawar:
Hey @ericpromislow looks like your pipeline has resumed to consuming our latest commits of our `ci-passed` branch already which was basically what this PR was attempting to do. This commit just went through on `develop` which is all I wanted to do with this PR: https://github.com/cloudfoundry/cf-for-k8s/commit/2554e6dc5a5b8ac96ae1ceed326abda7d116c683

Gonna close this out then. Thanks though!

PR #472, opened 2020-09-24
Make application wait for sidecar proxy before starting

### LINKED ISSUE #189: Starting a docker image app fails if it tries to establish connections before Envoy is ready
### Summary

Starting or restarting a docker image app fails if the app tries to establish connections before the Envoy proxy is ready, and this communication failure causes the app to crash.

```
$ cf restart service-manager
Restarting app service-manager in org SAP_servicemanager / space service-manager as admin...

Stopping app...

Waiting for app to start...
Start unsuccessful
```

Examining the corresponding pod shows that the `opi` container was restarted once due to an error. The logs from the first (failed) run are as follows:

```
{"component_type":"application","correlation_id":"-","level":"info","logger":"sm/sm.go:105","msg":"Setting up Service Manager storage...","type":"log","written_at":"2020-05-08T09:23:04.068924213Z","written_ts":"1588929784"}
panic: unable to construct service-manager builder: error opening storage: error opening storage: could not update database schema: dial tcp 100.65.158.99:5432: connect: connection refused

goroutine 1 [running]:
main.main()
	/go/src/github.wdf.sap.corp/SvcManager/sm-sap/main.go:34 +0x783
```

The above logs indicate that the app tried to connect to its database, but the connection was refused, which caused the app to crash. The reason that the connection was refused is that by that time (2020-05-08T09:23:04), the Envoy proxy hasn't started yet, since the `istio-proxy` logs indicate that this happens a second later:

```
2020-05-08T09:23:05.767841Z	info	Envoy proxy is ready
```

After the `opi` container is restarted by Kubernetes, the app starts successfully and connects to its database. So the end result is a running app, but a failed `cf start`.

This is reproducible with several apps on our side. All of them try to establish connections before Envoy has started. Some of them treat the "connection refused" error as fatal and crash, leading to the above error. Others treat it as non-fatal or retry, and so avoid the error.

### Reproduction Steps

Push an app from a docker image that is fairly quick to start, tries to establish a connection immediately after starting, and treats the failure as fatal. For best effect, push it with `--no-start` and then start it with `cf start`, or restart it with `cf restart`.

### Expected behavior

`cf start` or `cf restart` don't return an error in this case. This could be achieved by either ensuring that Envoy proxy starts before the app tries to establish a connection, or treating the first crash from the app as non-fatal.

mike1808:
## WHAT is this change about?
This change will cause application containers to not start until the `istio-proxy` container is not up. This will prevent issues when the application requires network connectivity right at the start.

One caveat about how this works:

>Something to note here is that this feature leverages an undocumented behaviour of the Kube runtime. The fact that the kubelet blocks the startup of subsequent containers is largely an unintended feature and happenstance.
>
>There have been efforts to change this core behaviour of Kubernetes:
>
>kubernetes/kubernetes#81450
>https://github.com/kubernetes/kubernetes/pull/82282/files
>kubernetes/kubernetes#86115
>
>So we must be aware that there is a possibility of this unintended feature being removed even before SIdecar container first-class support is added: kubernetes/enhancements#753

Closes #189

No

## Acceptance Steps
Deploy an app which requires network at the start. You can use [this simple go app](https://gist.github.com/lucaschimweg/2d1d93aa81b5a4f998ecbb66fa63572e). It should successfully start when pushed.

### DIFF
--- a/build/istio/istioctl-values.yaml
+++ b/build/istio/istioctl-values.yaml
@@ -79,4 +79,6 @@ spec:
         }
     global:
       controlPlaneSecurityEnabled: true
+      proxy:
+        holdApplicationUntilProxyStarts: true
 
--- a/config/istio/istio-generated/xxx-generated-istio.yaml
+++ b/config/istio/istio-generated/xxx-generated-istio.yaml
@@ -8300,7 +8300,7 @@ data:
           "excludeIPRanges": "",
           "excludeInboundPorts": "",
           "excludeOutboundPorts": "",
-          "holdApplicationUntilProxyStarts": false,
+          "holdApplicationUntilProxyStarts": true,
           "image": "proxyv2",
           "includeIPRanges": "*",
           "logLevel": "warning",


ericpromislow:
LGTM - retry after this moves out of Draft status

mike1808:
@ericpromislow I don't see CI links in the comments. Can you please retry?

ericpromislow:
Which CI links are you referring to? The CI tests are all passing. I already reviewed the PR. Should be merged and closed soon.

PR #473, opened 2020-09-24
Docs: Add information on scaling networking components

KauzClay:
## WHAT is this change about?
We are adding a doc to explain how operators might write an overlay to scale networking components. We include an example overlay file and encourage it's use.

We would appreciate your feedback on this. Is this what you are looking for, based on the conclusions in this [doc](https://docs.google.com/document/d/1VwOdsKXUhC47Xw7Cutj7Fx-NuYTSV-diZB8Qkt61xfw/edit#heading=h.jrqmnlxmgtza)?

[#174613651](https://www.pivotaltracker.com/story/show/174613651)

No

## Acceptance Steps
```gherkin
Given cf-for-k8s repo
When I look at platform operator docs
Then I see a document on how to scale networking components
```

@XanderStrike @cloudfoundry/cf-for-k8s-networking

### DIFF
--- a/docs/platform_operators/scaling-networking.md
+++ b/docs/platform_operators/scaling-networking.md
@@ -13,6 +13,9 @@ environment, simply:
 1. Adjust the values to your liking
 1. Append `-f scaling-networking.yml` to the ytt command you run for deployment
 
+The number of ingressgateway replicas depends on load profile of your cluster. 
+The number istiod replicas depends istiod replicas the number of application instances.
+
 ```
 #@ load("@ytt:overlay", "overlay")
 #@ load("@ytt:json", "json")
--- a/docs/platform_operators/scaling-networking.md
+++ b/docs/platform_operators/scaling-networking.md
@@ -1,0 +1,126 @@
+# Scaling Networking
+
+## Local Development
+If you would like to scale the networking components of cf-for-k8s to accommodate
+for local development on a laptop, simply deploy following the deployment docs.
+By default, cf-for-k8s is shipped for a kick-the-tires experience.
+
+## Larger Scale Deployments
+If you would like to scale up the networking components for a larger scale
+environment, simply:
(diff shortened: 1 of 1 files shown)


docs/platform_operators/scaling-networking.md
```diff
@@ -0,0 +1,136 @@
...
+#@ routecontroller_cpu_limit = "400m"
+#@ routecontroller_mem_request = "32Mi"
+#@ routecontroller_mem_limit = "1024Mi"
+
+#@ sidecar_cpu_request = "100m"
+#@ sidecar_cpu_limit = "2000m"
+#@ sidecar_mem_request = "128Mi"
+#@ sidecar_mem_limit = "1024Mi"
+
+#! DO NOT MODIFY BELOW THIS LINE
```
jamespollard8, line 54: maybe this should say something like "the intent is that the overlays below shouldn't need to be changed"

jamespollard8 (commented):
Thanks for this!

I believe we may have discussed this elsewhere (or on zoom) but I can't find the other thread at the moment.

Our current thinking is that we'll provide the overlay instructions for all components in one doc. We have that work prioritized here: [tracker story](https://www.pivotaltracker.com/story/show/174900829). As part of that work, we'll incorporate this documentation and I imagine the specific recommendations regarding scaling networking will live on from this PR.

Does that make sense?

XanderStrike:
Does that mean that this PR will be merged soon and y'all may modify this file in the future? That's totally fine. Or should we close this and you'll take these commits and roll them into your own PR?

jamespollard8:
> Or should we close this and you'll take these commits and roll them into your own PR?

let's do it this way. I added a task in [our story](https://www.pivotaltracker.com/story/show/174900829) to reference/steal some of the language and documentation here

Thanks again

PR #475, opened 2020-09-28
Rename allow-istio-control-plane network policy

rodolfo2488:
* We decided that renaming this network policy to
"only-allow-istio-control-plane" because it better describes what this
policy does. It is a policy that allows traffic from the istio to all
pods in a namespace.

[174810340](https://www.pivotaltracker.com/story/show/174810340)

Co-authored-by: Rodolfo Sanchez <srodolfo@vmware.com>

>

## Acceptance Steps
```
When I have a cluster with cf-for-k8s 
And I get the network policies in all namespaces `kubectl get netpol -A  | grep allow-istio-control-plane`
Then I can see network policies called only-allow-istio-control-plane.
And I do not see network policies called allow-istio-control-plane 
```

### DIFF
--- a/config/networking/network-policies.yaml
+++ b/config/networking/network-policies.yaml
@@ -71,7 +71,7 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: allow-istio-control-plane
+  name: only-allow-istio-control-plane
   namespace: #@ system_namespace()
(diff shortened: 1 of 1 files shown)


jamespollard8 (approved):
Sweet - looks good to me. Thanks @rodolfo2488!

The failing tests are unrelated and have been fixed since they ran here.

PR #478, opened 2020-09-29
Add smoke test bigger than 64k

modulo11:
## WHAT is this change about?
This change ensures that the app will be pushed to the blobstore as well. See https://github.com/cloudfoundry/cf-for-k8s/pull/434#issuecomment-695141395 for details.

No

## Acceptance Steps
Run smoke-test and see if the blobstore contains a blob.

@phil9909 
@loewenstein


tests/smoke/smoke_test.go
```diff
@@ -113,6 +114,33 @@ var _ = Describe("Smoke Tests", func() {
...
+
+		When("app source is < 64k", func() {
+			It("creates a routable app pod in Kubernetes from a source-based app", func() {
+				expectSuccessfulPush("assets/test-node-app")
+			})
+		})
+
+		When("app source is > 64k", func() {
+			var tempdir string
+			BeforeSuite(func() {
```
davewalter, line 127: Minor nit: I think this could/should be a `BeforeEach` instead of a `BeforeSuite`.

tests/smoke/smoke_test.go
```diff
@@ -113,6 +114,33 @@ var _ = Describe("Smoke Tests", func() {
...
+				bigFileName := path.Join(tempdir, "a-big-file.txt")
+				bigFileContent := []byte(strings.Repeat("na", 32*1024) + " batman")
+				Expect(ioutil.WriteFile(bigFileName, bigFileContent, 0o600)).To(Succeed())
+			})
+
+			It("creates a routable app pod in Kubernetes from a source-based app", func() {
+				expectSuccessfulPush(tempdir)
+			})
+
+			AfterSuite(func() {
```
davewalter, line 141: Minor nit: I think this could/should be an `AfterEach` instead of an `AfterSuite`.

davewalter (commented):
Hi @modulo11,

Looks good overall. I left a couple of minor comments.

Regards,
Dave

tests/smoke/smoke_test.go
phil9909: Why? Currently, there is only one test, so it won't make a difference. But if there are multiple `It`s in this `When`-block, then we would create (and delete) multiple (identical) temp directories when using `BeforeEach`. This seems unnecessary to me.

tests/smoke/smoke_test.go
ericpromislow: I agree with Dave's comments: I looked at 3 different CF test directories, and all of them use `BeforeSuite` and `AfterSuite` in top-level `suite_test` files only. Everywhere else, they use `BeforeEach` and `AfterEach`.  If you want the `Before...` actions to apply to only the one test, you can wrap it in a Context block. Typically, we avoid wrapping a single test in a Context block, and add them when more tests are added.

davewalter (commented):
Hi @modulo11,

Sorry for the delayed response. @Birdrock and I were just reviewing this again, and realised that this introduces a strange dependency between smoke-tests and the `run-external-blobstore-validation-test` CI task that attempts to validate the external blobstore configuration.

We think it would make more sense if the app that the `run-external-blobstore-validation-test` task pushes was > 64K as that would keep it self-contained. It would also mean that we don't run this unnecessary extra test on all other deployments.

Let us know what you think.

Dave and @Birdrock

modulo11:
Hi @davewalter,

I'm fine with that PR #434 already contains a test with an app >64k. I can rebase that branch and then we can continue. Afterwards I'd close #478.
What do you think?

acosta11:
Hi @modulo11 ,

Yeah that sounds great. Focusing on #434 now.

Thanks,
Andrew and @davewalter

davewalter:
Closing now that PR #434 has been merged.

PR #481, opened 2020-09-30
ENH: Separate "external" and "internal" configuration

davewalter:
## WHAT is this change about?
The goal of this change is to separate the "external" user-facing configuration from the "internal" intra-platform configuration required for CF for K8s to operate. It isolates the internal configuration in the `generate-internal-values.sh` script that will eventually be obsoleted by the implementation of a cluster-based secret management system (probably Quarks) and provides clear documentation and samples for the external configuration that the user is expected to provide when deploying cf-for-k8s.

One of my goals was to provide a working sample external values file for deploying locally on KinD, but that would have meant including real system and workloads certificates for vcap.me, so I decided against that.

No

## Acceptance Steps
Deploy and run smoke-tests.


paulcwarren:
My feeling is that this is good change.  

AFAICT it doesn't really change the operator experience for 1.0 either.  It remains something along the lines of:
- generate your values (just now it is internal values, rather than all values)
- kapp deploy with values (internal) + values (real - and for kind we provide a sample)

I would also note that this seems to play well with the decision to **not** switch all of our default values up to orient towards the local/developer deployment and instead leave them oriented towards a IaaS deployment.

jamespollard8 (changes requested):
>One of my goals was to provide a working sample external values file for deploying locally on KinD, but that would have meant including real system and workloads certificates for vcap.me, so I decided against that.

@ericpromislow, @paulcwarren and I are in favor of providing a real cert here, ideally with a 10 year expiration.

It'd be particularly useful for things like deploying an app_registry for the all local use case. (We've not been able to get this working with fake or self-signed certs)

jamespollard8:
I've temporarily set the pipeline to use ci from this branch to test itself. The tests are re-running now

ericpromislow:
Needs to be rebased against develop if we're going to test and fly this manually.

davewalter:
Recreated as #506 to resolve conflicts.

Issue #460, opened 2020-09-18
Input values for certificates should have a consistent format


julian-hj:
**Is your feature request related to a problem? Please describe.**
Some user provided certificates are base64 encoded, such as the ingress certs. Others are not and we base64 encode them later. This is confusing to users.

**Describe the solution you'd like**

It would be better if we could make the interface we provide to the operator consistent--either all encoded, or all not encoded.

### A/C
When I follow directions to add certificates to my configuration input
Then I can use the same format for all of my certificates

emalm:
I'd advocate for PEM-encoded, but not base64-encoded, for any certificate or RSA private key input values. Platform engineers typically receive PEM-encoded representations of these values and should not have to bother with the extra step of base64-encoding these values when either the YTT templates or the K8s API itself can do it for them. It's also harder to identify a formatting or omission error in a single line that is thousands of characters long than it is in a PEM-encoded block of text.

davewalter:
Fixed in #482.

PR #482, opened 2020-09-30
ENH: Allow operators to provide PEM-encoded certs

davewalter:
## WHAT is this change about?
This change allows operators to provide PEM-encoded system and workloads certificates that are not base64 encoded. It moves the bas64 encoding logic into the templates where the values are provided to components. It also cleans up some cruft in the generated values.yml file. This PR addresses issue #460.

Yes

## Acceptance Steps
Generate a config values file with system and workloads certificates that are NOT base64 encoded, deploy and run smoke-tests.

### DIFF
--- a/config/eirini/eirini.yml
+++ b/config/eirini/eirini.yml
@@ -17,9 +17,9 @@ metadata:
   name: eirini-internal-tls-certs
   namespace: #@ system_namespace()
 data:
-  tls.crt: #@ data.values.internal_certificate.crt
-  tls.key: #@ data.values.internal_certificate.key
-  tls.ca: #@ data.values.internal_certificate.ca
+  tls.crt: #@ base64.encode(data.values.internal_certificate.crt)
+  tls.key: #@ base64.encode(data.values.internal_certificate.key)
+  tls.ca: #@ base64.encode(data.values.internal_certificate.ca)
 
 #! Allow app traffic from the istio-ingressgateway
 ---
--- a/config/eirini/eirini.yml
+++ b/config/eirini/eirini.yml
@@ -16,10 +16,10 @@ kind: Secret
 metadata:
   name: eirini-internal-tls-certs
   namespace: #@ system_namespace()
-data:
-  tls.crt: #@ base64.encode(data.values.internal_certificate.crt)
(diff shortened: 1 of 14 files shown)


config/eirini/eirini.yml
```diff
@@ -17,9 +17,9 @@ metadata:
   name: eirini-internal-tls-certs
   namespace: #@ system_namespace()
 data:
```
davewalter, line 19: Can we use stringData here and let K8s handle the base64 encoding?

config/istio/external-routing.yml
```diff
@@ -12,8 +13,8 @@ metadata:
   annotations:
     kapp.k14s.io/change-rule.istio-ingressgateway: "upsert before upserting istio.io/ingressgateway"
 data:
```
davewalter, line 15: Can we use stringData here and let K8s handle the base64 encoding?

config/istio/external-routing.yml
```diff
@@ -24,8 +25,8 @@ metadata:
...
```
davewalter, line 27: Can we use stringData here and let K8s handle the base64 encoding?

config/uaa/uaa.yml
```diff
@@ -247,8 +248,8 @@ metadata:
   namespace: #@ system_namespace()
 type: Opaque
 data:
```
davewalter, line 250: Can we use stringData here and let K8s handle the base64 encoding?

davewalter (commented):
Review with Birdrock. See inline comments.

tests/ytt/uaa/uaa-values.yml
```diff
@@ -8,8 +8,8 @@ data:
   uaa.yml:
 
 internal_certificate:
-  key:
-  crt:
+  key: internal-cert-key
```
paulcwarren, line 11: These looks related/unrelated to me.  How does moving where base64 encoding mean that we now need to pass in a value for key and crt here?

paulcwarren:
In general, this looks like an excellent changeset to me.

jamespollard8 (approved):
Looks good to me and @ericpromislow - thanks @davewalter!

PR #485, opened 2020-09-30
FIX: disable_snapshot_annotations to fix #444

### LINKED ISSUE #444: `cf push` fails on KinD: cf-default-builder image "labels" size greater than max allowed size
duplicate of https://github.com/pivotal/kpack/issues/473

## Describe the bug
We've been seeing consistent failures in our CI when running smoke tests on KinD. After debugging, we found this image pull failure:
`Failed to pull image "gcr.io/.../cf-default-builder@sha256:977feb...": rpc error: code = InvalidArgument desc = failed to pull and unpack image "gcr.io/cf-relint-greengrass/cf-workloads/cf-default-builder@sha256:977febd...": failed to prepare extraction snapshot "extract-712270633-CRQK sha256:f0dcba2...": info.Labels: label key and value greater than maximum size (4096 bytes), key: containerd: invalid argument`

From the smoke test output, this manifests as hanging after these lines:
```
Staging app and tracing logs...
   Loading secret for "gcr.io" from secret "cc-kpack-registry-auth-secret-ver-1" at location "/var/build-secrets/cc-kpack-registry-auth-secret-ver-1"
   Successfully downloaded cf-blobstore-minio.cf-blobstore.svc.cluster.local:9000/cc-packages/d7/d0/d7d01fef-dfdd-44df-8a3b-a03e40379c29 in path "/workspace"
... <- hangs here
```

## To Reproduce*
1. Setup a KinD cluster with Kubernetes version 1.18.4 or later and connect to it
2. Follow the docs to install cf-for-k8s on kind https://github.com/cloudfoundry/cf-for-k8s/blob/master/docs/deploy-local.md#steps-to-deploy-on-kind
3. Run smoke tests or `cf push`

## Expected behavior
Smoke tests succeed on KinD.

## Additional context
As a work-around, we've added an overlay to tear out all non-node buildpacks from the cf-default-builder. 59fa844

jamespollard8:
Fixes #444 

## WHAT is this change about?
The issue is described thoroughly in #444. This fix idea came from bentheelder in kubernetes slack #kind channel here: [thread](https://kubernetes.slack.com/archives/CEKK1KTN2/p1601490158071400?thread_ts=1601490068.071200&cid=CEKK1KTN2)

Nope

## Acceptance Steps
KinD should deploy without using the `only-nodejs-builder-overlay.yml` hack/work-around.

To test this properly in CI, I've temporarily pointed CI to use this branch for the `create-kind-cluster` and `install-cf-on-kind` tasks.


jamespollard8:
On second thought, because
1) I've tested this locally on kind with k8s versions 1.16 and 1.19 
and
2) testing this in a PR requires this CI work-around business

I think it'll make more sense to just have this tested as part of our main pipeline after merging to develop (which also tests kind, but will automatically use the relevant CI changes from this PR)

PR #495, opened 2020-10-01
add option to allow Prometheus scrape access to metrics endpoints

JamesClonk:
## WHAT is this change about?
Adds an optional flag `allow_prometheus_metrics_access` to allow any Prometheus server (or other component) access to the metrics endpoint of cf-system components.

If a user already has a Prometheus server installed on his K8s, as might be common via helm chart _stable/prometheus_ for example, this Prometheus server would not be able to scrape the Prometheus `/metrics` endpoints of uaa, cf-api-server and metrics-proxy in their current configuration. The Istio-proxy present within the pods restricts access towards the metrics endpoints for any external client.
This change introduces a new flag with ytt overlays that add an additional Istio annotation onto the pods to exclude the metrics endpoints from having their access restricted by Istio.

Yes

## Acceptance Steps
Generate deployment manifest with ytt with and without `allow_prometheus_metrics_access` being set, verify additional Istio annotations being present in resulting output.


jamespollard8 (changes requested):
Sweet - looks good @JamesClonk 

Would you be willing/able to add a couple unit tests for this? You should be able to start from these: https://github.com/cloudfoundry/cf-for-k8s/blob/005969c52999a030ecadfa3d47b26e1cb3ec7c13/tests/ytt/external_db_test.go#L26-L66

If you need any new yttk8smatchers, it's easy to get those added to https://github.com/cloudfoundry/yttk8smatchers

JamesClonk:
> Sweet - looks good @JamesClonk
> 
> Would you be willing/able to add a couple unit tests for this? You should be able to start from these:
> 
> If you need any new yttk8smatchers, it's easy to get those added to https://github.com/cloudfoundry/yttk8smatchers

@jamespollard8 I've added unit tests to check for the Istio rule.

jamespollard8 (approved):
Looks great - thanks for adding those tests!

kkburr:
@JamesClonk Hi there, I'm on the team that's working on getting prometheus out of the box in cf-for-k8s. Can you give me a little more information about how you're deploying prometheus when we use this? Specifically, is it via helm? Which namespace is prometheus in?

We have a prometheus deployment [here](https://github.com/cloudfoundry/cf-k8s-prometheus) that is meant to scrape securely, but it might be blocked by a bug in the 1.7.1 istio release that got merged into cf-for-k8s recently. I'm looking to see what our options are in terms of prometheus recommendations for this first cf-for-k8s release.

JamesClonk:
Hi @kkburr 
We've had a few test installations before with the `stable/prometheus` Helm chart, but nowadays are mainly using the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator). All our Prometheus server instances are deployed via this operator and we additionally add network-policies to allow traffic where necessary. While there's usually a Prometheus instance in the namespace `prometheus`, the namespace could potentially be named anything. 😅 

Using https://github.com/cloudfoundry/cf-k8s-prometheus is fine I guess for K8s clusters where there isn't already some other form of Prometheus present, but I suspect that it's not just our company who already has an existing Prometheus infrastructure and thus cannot make use of cf-k8s-prometheus.

That's odd, I rerun ytt just now and can see the annotations on cf-api-server, uaa and metric-proxy. We wouldn't have any of their metrics in our Prometheus if it didn't work. 🤔 
Are you sure you set it to true in your cf-values.yml and reran the ytt templating?

PR #497, opened 2020-10-01
Use the capi-database-encryption-key-secret in capi-k8s-release

matt-royal:
Use the capi-database-encryption-key-secret in capi to encrypt the database. Formerly we used "TODO"

NOTE: this commit manually bumps to a newer version of capi-k8s-release. Without this manual bump, this PR would cause cf-for-k8s to fail to deploy (due to ytt overlay issues)

NO

## Acceptance Steps
cf-for-k8s should continue to deploy successfully. There's no external test for the db encryption key. This is a chore on the CAPI/KAT/CAKE side.

@njbennett 

Here's our story for this work: https://www.pivotaltracker.com/story/show/175080272

### DIFF
--- a/config/capi/_ytt_lib/capi-k8s-release/templates/secrets.lib.yml
+++ b/config/capi/_ytt_lib/capi-k8s-release/templates/secrets.lib.yml
@@ -10,6 +10,9 @@
 - name: cf-api-db-password
   secret:
     secretName: #@ data.values.ccdb.password_secret_name
+- name: cf-api-db-encryption-key
+  secret:
+    secretName: #@ data.values.ccdb.encryption_key_secret_name
 - name: cf-api-username-lookup-client-secret
   secret:
     secretName: #@ data.values.uaa.clients.cloud_controller_username_lookup.secret_name
@@ -22,7 +25,8 @@
 
 #@ ccng_config_mount_path = "/config/cloud_controller_ng.yml"
 #@ ccng_secrets_mount_path = "/config/secrets.yml"
-#@ ccdb_mount_path = "/etc/cf-api/ccdb"
+#@ ccdb_password_mount_path = "/etc/cf-api/ccdb_password"
+#@ ccdb_encryption_key_mount_path = "/etc/cf-api/ccdb_encryption_key"
 #@ cf_api_username_lookup_client_secret_mount_path = "/etc/cf-api/username-lookup-client-secret"
 #@ cf_api_blobstore_secret_access_key_mount_path = "/etc/cf-api/blobstore/secret-access-key"
 
@@ -40,7 +44,10 @@
 #! secrets for credentials used in config
 
 - name: cf-api-db-password
-  mountPath: #@ ccdb_mount_path
+  mountPath: #@ ccdb_password_mount_path
+
+- name: cf-api-db-encryption-key
+  mountPath: #@ ccdb_encryption_key_mount_path
 
 - name: cf-api-username-lookup-client-secret
   mountPath: #@ cf_api_username_lookup_client_secret_mount_path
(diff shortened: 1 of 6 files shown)


matt-royal:
The failing test seems like a flake. How can I rerun it?

config/capi/_ytt_lib/capi-k8s-release/values/_default.yml
```diff
@@ -18,6 +18,7 @@ ccdb:
   database: cloud_controller
   host: capi-database-postgresql
   password_secret_name:
+  encryption_key_secret_name: capi-database-encryption-key-secret
```
cwlbraa, line 21: this shouldn't block merging, but why do we configure these in cf-for-k8s when we could hardcode them in capi-k8s-release? I guess it gives some decoupling between the two where we could ostensibly deploy capi-k8s-release on its own, but the coupling is still apparent in the fact we have to make this PR at all...

cwlbraa (commented):
just want to call out to cf-for-k8s maintainers that this change is critical for 1.0.

ericpromislow:
The way that `verify-existing-app` failed after the upgrade, I would bet that CC ran into a mis-encrypted field.
It just gives an `An unknown error occurred.` error message, and I didn't see likely logs in either of the cf-api-servers. 
But the CF was deleted before I could investigate further.

config/capi/_ytt_lib/capi-k8s-release/values/_default.yml
matt-royal: The rationale here is that the secret is defined in cf-for-k8s, so it's the source of truth for the name. We decided this time to give the value in capi-k8s-release a default so it wouldn't break our pipeline while we waited for this PR to merge. We should be able to safely take out the default value after this is merged.

Another option we have is that we could move the Secret definition into capi-k8s-release and avoid the coupling entirely. I'm not sure this meshes with the long term plan for secrets in cf-for-k8s, though.

matt-royal:
On further reflection, the upgrade failure is unsurprising. The upgrade changes the db encryption key and we don't expect the CF API to continue to work afterwards. Given that, I'd say we should ignore the failure.

tcdowney:
> In other words, this changes requires a fresh install.

This is one of the reasons it's critical this gets in before 1.0. 🙂

config/capi/_ytt_lib/capi-k8s-release/values/_default.yml
cwlbraa: I do now recall this being prescribed by the secrets proposal. With time and distance, I still remember that there are good reasons to have the secrets in cf-for-k8s (they're part of stable user configuration) but I'm not sure why the name needs to be configurable on the capi-k8s-release values interface.

config/capi/_ytt_lib/capi-k8s-release/templates/secrets.lib.yml
```diff
@@ -22,7 +25,8 @@
 
 #@ ccng_config_mount_path = "/config/cloud_controller_ng.yml"
 #@ ccng_secrets_mount_path = "/config/secrets.yml"
-#@ ccdb_mount_path = "/etc/cf-api/ccdb"
+#@ ccdb_password_mount_path = "/etc/cf-api/ccdb_password"
+#@ ccdb_encryption_key_mount_path = "/etc/cf-api/ccdb_encryption_key"
```
ericpromislow, line 29: Why are there no extensions on the new files' paths? Is it because they're simple plain text files?

ericpromislow:
Looks like this comment got deleted by a github glitch:

[finishes #175081868](https://www.pivotaltracker.com/story/show/175081868)

Co-authored-by: rchu@pivotal.io

config/capi/_ytt_lib/capi-k8s-release/templates/secrets.lib.yml
matt-royal: @ericpromislow They're directories, with files inside for each different field. In this case the only file is `password`, which we add to the variable later in this same yaml file.

ericpromislow: right, thx

config/capi/_ytt_lib/capi-k8s-release/values/_default.yml
reneighbor: @matt-royal Eric and I had the same Q-- why is the name of the secret configurable?

PR #507, opened 2020-10-04
Added base64 instructions and expectations

ryandotclair:
>

## WHAT is this change about?
Added clearer instructions on expected values of the Let's Encrypt certs (base64). Not everyone might be familiar with this behavior.

No

## Acceptance Steps
N/A?

N/A

## Things to remember
- Our CI uses concourse task files from HEAD of the `develop` branch. If your PR includes/requires CI changes, please ping the RelInt interrupt and they'll help apply the CI changes for you.
- Checkout the [contributing guidelines](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/community/CONTRIBUTING.md)

### DIFF
--- a/docs/platform_operators/setup-ingress-certs-with-letsencrypt.md
+++ b/docs/platform_operators/setup-ingress-certs-with-letsencrypt.md
@@ -60,15 +60,17 @@ _acme-challenge.$APPS_DOMAIN.	TXT    kyfxzsAirB79lsk173jkdlamxiryqloy
 ```
 
 ### Update cf-values yaml
-The following instructions assume you have created `cf-install-values.yml`. Please ensure to copy the file contents into the variables as is.
+The following instructions assume you have created `cf-install-values.yml`. Please ensure to base64 the 4 files that were created in above steps into the variables below.
+
+> Example of how to base64: `cat /tmp/certbot/cfg/live/$SYS_DOMAIN/fullchain.pem | base64`
 
 1. **Update system certificate values**
 
     Lookup `system_certificate` in `cf-install-values.yml`. You should config variables `crt`, `key` and `ca`. Follow the instructions below,
     ```yaml
     system_certificate:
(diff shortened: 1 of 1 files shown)


davewalter:
Hi @ryandotclair, Thanks for the contribution. I don't think this PR makes sense, given that we recently merged PR #482. Users don't need to base64-encode the certificates in their values.yml file anymore.

ryandotclair:
@davewalter oh cool. So how does that need to look in the documentation then? Just the raw text (strip out the BEGIN/END lines)? I'd tried that method and it didn't work, but could just be my particular build. Net-Net, I still think documentation needs to change to be more verbose.

paulcwarren:
ryandotclair, you just need to copy the certificate provided by letsencrypt as-is.  ----BEGIN CERTIFICATE---- to ----END CERTIFICATE---- inclusive and you can leave carriage returns and spaces.  (or equivalent for the pk).  Indented appropriately within the yaml structure.  (Note, if you don't indent appropriately then you get ytt parsing errors I think so hopefully this is obvious).

Or at least that is what I did yesterday when following the docs.  That's what seemed like the natural thing to do to me.  But please let us know if you think different.  You viewpoint is more valid than mine as I am a lot closer to the project everyday.

What do you think?

ryandotclair:
That's pretty cool. The hangup for me as a new user is the example provided (as the only reference point for me) and the "replace this with contents of file" verbiage just prior to that:

image
 

It's not obvious to me what that is, since that's not the contents of the file (read: no BEGIN mentioned).

davewalter:
@ryandotclair I think this was a miss on my part when I implemented this change in PR #482. I thought I had updated the docs appropriately, but missed this section of the `setup-ingress-certs-with-letsencrypt.md` doc. I submitted PR #519 to hopefully address this. Please let me know if this change helps clarify the documentation.

jamespollard8:
Closing in favor of #519

ryandotclair:
Perfect. Thanks @davewalter!

PR #512, opened 2020-10-06
Disable prometheus metrics merging

XanderStrike:
[#175148731](https://www.pivotaltracker.com/story/show/175148731)

Co-authored-by: Mikael Manukyan <mmanukyan@vmware.com>

## WHAT is this change about?

Removes the merging of sidecar and workload stats, the metrics team will scrape each container independently.

- this caused some problems for the metrics team
- namely, it causes metrics to be emitted in plaintext

Nope

## Acceptance Steps
N/A

@mike1808 @cloudfoundry/cf-for-k8s-networking

### DIFF
--- a/build/istio/istioctl-values.yaml
+++ b/build/istio/istioctl-values.yaml
@@ -40,7 +40,7 @@ spec:
       rewriteAppHTTPProbe: true
     meshConfig:
       accessLogEncoding: 'JSON'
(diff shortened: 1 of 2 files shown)


XanderStrike:
don't judge me i want that shirt

XanderStrike:
The external db failure is not related to this change.

PR #487, opened 2020-09-30
Use full paketo build and run bionic stack images

paulcwarren:
[#175053873](https://www.pivotaltracker.com/story/show/175053873)

PR #513, opened 2020-10-06
Fork and patch Eirini to loosen UID requirement

Birdrock:
## WHAT is this change about?
This PR configures cf-for-k8s to use a forked and patched version of Eirini images that loosens the UID requirement. It removes the UID 2000 requirement, while still requiring non-root through the security context

No.

## Acceptance Steps
Deploy with a stack that has a non-2000 (non-root) UID. The reference set we used is contained in this PR: https://github.com/cloudfoundry/cf-for-k8s/pull/487

@paulcwarren @acosta11


mike1808:
We're getting `Error: container has runAsNonRoot and image will run as root` when trying to deploy a docker app after this change.

PR #516, opened 2020-10-07
reformat version in upgrade job

KauzClay:
Co-authored-by: Kauana dos Santos <kdossantos@vmware.com>

## WHAT is this change about?
We want to avoid having `.`s in pod names because some platforms don't
allow that.

[#175169873](https://www.pivotaltracker.com/story/show/175169873)

No

## Acceptance Steps
**When** I have a cluster with cf-for-k8s deployed
**And** I describe the `restart-sidecar-job` with `k get po -n cf-workloads | grep "restart"`
**Then** I see that the pod name is `restart-workloads-for-istio1-7-1-<hash>` and **not** `restart-workloads-for-istio1.7.1-<hash>`

@kauana @cloudfoundry/cf-for-k8s-networking

### DIFF
--- a/build/istio/build.sh
+++ b/build/istio/build.sh
@@ -7,4 +7,4 @@ ${SCRIPT_DIR}/generate.sh "$@" | kbld -f - > "${SCRIPT_DIR}/../../config/istio/i
 
 # save the current Istio version in the networking configs
 ISTIO_VERSION="$(< "${SCRIPT_DIR}/values.yaml" yq -r .istio_version)"
-sed -r -i'' 's/(return)(.*)$/\1 "'${ISTIO_VERSION}'"/' "${SCRIPT_DIR}/../../config/istio/upgrade-istio-sidecars-job.yml"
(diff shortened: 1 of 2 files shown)


ericpromislow:
The `check-vendir-sync` job failed because we run `build/istio/build.sh` after running `vendir sync`.

In particular, https://github.com/cloudfoundry/cf-for-k8s/blob/db6f0aa9a6db96ba43b0249fdf11adeb83a8fc32/build/istio/build.sh#L9-L10 rewrites the version field

This replaces whatever follows the `return` stmt in `config/istio/upgrade-istio-sidecars-job.yml`
with the dotted version.

KauzClay:
@ericpromislow Yeah, that sed command was so we could change istio version in one place and have our tooling replace it in all the necessary places.

We changed the regex in the sed to be more specific, and tweaked the overlay. I think it should work now.

davewalter:
For historical record, the rename of the `restart-workloads-for-istio1-7-3` job caused app downtime during the upgrade. This was expected as all application pods were restarted as part of the upgrade.

Issue #365, opened 2020-08-10
Request: air-gapped environment image relocation guide


acosta11:
**Is your feature request related to a problem? Please describe.**
As a user of cf-for-k8s with heightened security and availability requirements on my image registry, I would like to control the location of the images used by cf-for-k8s. Currently, image locations are not fully configurable in the cf-for-k8s data values interface.

**Describe the solution you'd like**
I would like a quick-start guide for configuring references to all of the artifacts required to deploy cf-for-k8s. This would enable the creation of an air-gapped environment alongside networking policies.

Similarly, we would like ongoing support/validation of this guide from the cf-for-k8s maintainers.

**Describe alternatives you've considered**
We have a workaround for managing image references with the `kbld` tool from [this slack conversation](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1595953719061300).

**Additional context**
While `kbld` technically solves the image reference problem at the moment, it requires some manual toil of managing configuration files that point to all of the images that require relocation. The configuration of those images may change in the absence of process/validation from the maintainers of cf-for-k8s.

JamesClonk:
Also, will there be `offline` buildpacks available like in CF-for-BOSH? Those are a must have for an air-gapped/offline environment too.

PR #517, opened 2020-10-08
Add private-registry instructions

JamesClonk:
## WHAT is this change about?
Adds documentation on how to relocate all upstream images into a private registry.

Yes

## Acceptance Steps
Follow image relocation steps as describe in docs/platform_operators/private-registry.md, deploy cf-for-k8s


JamesClonk:
PR for https://github.com/cloudfoundry/cf-for-k8s/issues/365

config/values/00-values.yml
```diff
@@ -89,3 +89,10 @@ blobstore:
   resource_directory_key: "cc-resources"
   buildpack_directory_key: "cc-buildpacks"
   aws_signature_version: "2"
+
+#! configuration for private registry
+private_registry:
```
davewalter, line 94: Could you please rename this to `system_registry`?

sample-cf-install-values.yml
```diff
@@ -113,3 +113,11 @@ use_first_party_jwt_tokens: true
...
 load_balancer:
   enable: false
+
+#! optional configuration for using a private registry that has all cf-for-k8s images relocated to.
+#! see <docs/platform_operators/private-registry.md>
+private_registry:
+  add_image_pull_secrets: false
+  hostname: <registry_hostname>
+  username: <registry_username>
+  password: <registry_password>
```
davewalter, line 123: Could you please remove this change? We are still trying to figure out exactly how we want to position our sample values file, but the current feeling is that it does not need to include all available configuration options, as users can refer to the `config/values/00-values.yml` file for that.

config/private-registry.yml
```diff
@@ -0,0 +1,71 @@
...
+#@overlay/match by=overlay.subset({"kind": "Job"}), expects="0+"
+---
+spec:
+  template:
+    spec:
+      #@overlay/match missing_ok=True
+      imagePullSecrets:
+      #@overlay/append
+      - name: private-registry-auth-secret
+#@ end
```
davewalter, line 71: Could these be collapsed into a single overlay using the `overlay.or_op()` [helper](https://github.com/k14s/ytt/blob/develop/docs/lang-ref-ytt-overlay.md#overlayor_op)?

davewalter (changes requested):
Hi @JamesClonk,

Thanks for the contribution! The timing couldn't be better with Dockerhub starting to rate limit image pulls for unauthenticated clients. We think this change could actually be more widely helpful to avoid deployment issues when using the "standard" image locations.

With that in mind, we have made some suggestions for changes. We did test that everything deploys successfully when we include a paid dockerhub account in the `private_registry` configuration settings. It would also be nice if we could include a doc reflecting this usage as part of this change.

The other thing that would be great is test coverage in our `tests/ytt` suite. These would validate that the secrets were being properly injected into the correct namespaces, and that the overlays were making the desired changes to the other resources.

Regards,
Dave and @Birdrock

config/private-registry.yml
JamesClonk: Oh nice, I didn't know about `overlay.or_op()`. 
Yes of course, changed it. 👍

sample-cf-install-values.yml
JamesClonk: done

config/values/00-values.yml
JamesClonk: done

JamesClonk:
@davewalter How is `docs/platform_operators/system-registry.md` for explaining the usage? (Unfortunately writing documentation isn't exactly my strong suit 😂🙈)

Working on `tests/ytt` atm..

JamesClonk:
Added a basic ytt test for imagePullSecrets.

acosta11:
Hello,

Thanks for the contribution! The added tests look good. We're going to merge this and do a fast-follow on docs for some nits that wouldn't be worth blocking the merge on here.

Thanks,
Andrew @Birdrock and @jamespollard8

PR #519, opened 2020-10-08
DOC: Update setup-ingress-certs-with-letsencrypt

davewalter:
## WHAT is this change about?
Update the Let's Encrypt doc with corrected examples now that ingress certificates do not need to be base64 encoded.

No

## Acceptance Steps
Validate that the docs are up to date.


jamespollard8 (approved):
LGTM!

PR #525, opened 2020-10-14
Replace VirtualService with Ingress for routing to system components

kauana:
By removing the system component's dependency on Istio Virtual Services and using Ingress, will be able to support multiple Ingress solutions (e.g. Contour), with the Ingress API which is leveraged by both Istio and Contour.

 [#175234234](https://www.pivotaltracker.com/story/show/175234234)

## Acceptance Steps
When I have a cluster deployed with this PR
And I get the ingresses with `kubectl get -n cf-system ingress`
Then I see uaa, and capi Ingress resources called `capi-external` and `uaa-external`
And when I get Virtual Services with `kubectl get -n cf-system vs`
Then I only see `log-cache` Virtual Service*
**And** I can push apps and see logs

* cf-k8s-logging will remove their VirtualService once this PR is merged ([PR to replace log-cache's VS](https://github.com/cloudfoundry/cf-k8s-logging/pull/47))


mike1808:
@davewalter `concourse-ci/run-unit-tests ` works locally, but fails in CI. How we can rerun the CI test?

PR #538, opened 2020-10-16
175210100 update secret rotation branch

ericpromislow:
## WHAT is this change about?

Document which fields currently are not correctly rotated, and point to bugs.

Additionally, point out which fields are rotated, but cause a CI job to fail when they're changed due to the way the job is written.

No

## Acceptance Steps
Mostly, verify that the updater jobs in CI, like `cf-for-k8s-stability-tests/validate-values-rotation-redeploy` passes. 

Note that this is mostly a documentation change. It can be verified by changing one of the fields listed in [the rotation doc](https://github.com/cloudfoundry/cf-for-k8s/blob/175210100-update-secret-rotation-branch/docs/platform_operators/rotating-secrets-and-certs.md#specific-issues), redeploy, and verify the foundation still works.

@jamespollard8 

## Things to remember
[x] - Make sure this PR is based off the `develop` branch
[x] - Include any links to related PRs, issues, stories, slack discussions, etc... that will help establish context.
-- None

### DIFF
--- a/docs/platform_operators/rotating-secrets-and-certs.md
+++ b/docs/platform_operators/rotating-secrets-and-certs.md
@@ -42,8 +42,6 @@ use either the old password or the new one, and will fail.
 
 * `cf_admin_password` - manual upgrade works, CI/upgrade fails
 
-* `cf_db.admin_password` - manual upgrade works, CI/upgrade works
-
 * `uaa.database.password` - manual upgrade works, CI/upgrade fails: "Error unmarshalling the following into a cloud controller error: no healthy upstream"
 
 ## Notes on Currently Supported Rotations
--- a/docs/platform_operators/rotating-secrets-and-certs.md
+++ b/docs/platform_operators/rotating-secrets-and-certs.md
@@ -2,7 +2,35 @@
 Most secrets in cf-for-k8s can be rotated by simply changing the values in your `cf-values.yml` file and running a standard deploy using ytt and kapp. The rotation is complete when the kapp deploy succeeds.
 
 ## Exceptions
-As of September 2020, it is currently not possible to rotate `app_registry` credentials. Future work on this will be tracked by https://www.pivotaltracker.com/story/show/173090115
+
+The following fields currently cannot be rotated:
+
+* `blobstore.secret_access_key`
+* `capi.database.encryption_key`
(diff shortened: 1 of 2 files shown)


docs/platform_operators/rotating-secrets-and-certs.md
```diff
@@ -2,7 +2,54 @@
...
 
 ## Exceptions
-As of September 2020, it is currently not possible to rotate `app_registry` credentials. Future work on this will be tracked by https://www.pivotaltracker.com/story/show/173090115
+
+The following fields currently cannot be rotated:
+
+* `blobstore.secret_access_key`
+* `capi.database.encryption_key`
+* `capi.database.password`
+* `cf_admin_password`
```
jamespollard8, line 11: this one can be rotated, right?

acosta11: Do we want to update the `## Exceptions` section in general to indicate the context in which these rotations fail? It seems like a subset of the rotations are failing for uptimer validations but actually work manually?

docs/platform_operators/rotating-secrets-and-certs.md
```diff
@@ -2,7 +2,54 @@
...
+* `cf_db.admin_password`
+* `uaa.database.password`
+
+For example, rotating the cloud controller db encryption key is a breaking change. Rotating the key will
```
acosta11, line 15: Sorry about the late question here given the scope of the story should have been limited to current documentation, but did we happen to validate that the uaa encryption key is rotatable as well?

ericpromislow: It wasn't in the list of non-rotated keys emitted by `upgrade-cf-new-values` at https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s-stability-tests/jobs/validate-values-rotation-redeploy/builds/136

ericpromislow: Part of the confusion here is that every field has two names: the dotted name from the yaml files, and then the name used internally in the post-ytt yaml. Users will be interacting mostly with the dotted name, but there could be cases where they need to control the actual field name.

PR #541, opened 2020-10-19
Fix yq syntax in deploy doc

paulczar:

## WHAT is this change about?

[yq syntax](https://mikefarah.gitbook.io/yq/commands/read) was wrong in the docs 

no 

## Acceptance Steps

validate the YQ command as specified.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -205,7 +205,7 @@ Currently, we test the following two container registries:
 
    ```console
    cf auth admin <cf-values.yml.cf-admin_password>
-   # or using yq: cf auth admin "$(yq -r '.cf_admin_password' ${TMP_DIR}/cf-values.yml)"
+   # or using yq: cf auth admin "$(yq r ${TMP_DIR}/cf-values.yml 'cf_admin_password')"
    ```
 
 1. Create an org/space for your app:


docs/deploy.md
```diff
@@ -205,7 +205,7 @@ Currently, we test the following two container registries:
 
    ```console
    cf auth admin <cf-values.yml.cf-admin_password>
-   # or using yq: cf auth admin "$(yq -r '.cf_admin_password' ${TMP_DIR}/cf-values.yml)"
+   # or using yq: cf auth admin "$(yq r ${TMP_DIR}/cf-values.yml 'cf_admin_password')"
```
jamespollard8, line 208: interesting - this change actually makes this yq call go from working to not working for me.

I've installed `yq` via `brew install python-yq`, which `yq` are you using?

paulczar: ooooh interesting.  we have multiple versions of yq !?!?!

I use https://mikefarah.gitbook.io/yq/ (on linux v 3.1.1 ).  do you have a pre-3 version?  I think it may have had a more jq like syntax back then.

paulczar: If version mismatches like this exist, then maybe it would be better to skip it?  or use `cat tmp/cf-values.yml | grep '^cf_admin_password:' | awk -F': ' '{ print $2}'` which is [probably] more universal.

Birdrock: @paulczar There are two main flavors of `yq`. One of them is [kislyuk/yq](https://github.com/kislyuk/yq) managed via Python `pip` and the other is [mikefarah/yq](https://github.com/mikefarah/yq) - a binary compiled from Golang. Unfortunately they use significantly different syntax. We should definitely be clear which `yq` we are using in any given context.

acosta11:
I may have missed this when I closed #546 earlier this week. Given the other instances of yq sprinkled through the docs, I've added a separate story in the backlog to fix the docs wholesale.

https://www.pivotaltracker.com/story/show/175418841

Thanks,
Andrew

jamespollard8:
FWIW I think we should align on `python-yq` in our repo since that's what our CI is using and because it would make this change's blast radius smaller.

jamespollard8:
Thanks @paulczar for bringing this to our attention. We're going to align our docs on the python yq - as captured by the story Costa referenced above 👍

PR #547, opened 2020-10-21
Update ytt tests to better reflect operator worflow

davewalter:
## WHAT is this change about?
This PR updates the ytt tests to better reflect the operator workflow of providing a single "config" directory to ytt. It does this by copying the templates to be tested to a temporary directory that is provided to ytt. This allows library loading within the templates to use a consistent path structure that works within the tests and at deploy-time.

[#175362433](https://www.pivotaltracker.com/story/show/175362433)

No

## Acceptance Steps
Tests should pass.


tests/ytt/external_blobstore_test.go
```diff
@@ -1,30 +1,48 @@
...
 	})
 
 	JustBeforeEach(func() {
-		ctx = NewRenderingContext(templates...).WithData(data)
+		targetDir, err = ioutil.TempDir("", "")
+		Expect(err).NotTo(HaveOccurred())
+
+		ctx = NewRenderingContext(targetDir, valueFiles...).WithData(data)
+
+		err = ctx.CopyTemplatesToTargetDir(templateFiles...)
```
paulcwarren, line 36: Hmmm...purely a suggestion and probably a nit pick but I wonder if a slightly better API might be?

```
ctx = NewRenderingContext(targetDir, valueFiles...).WithData(data).WithTemplateFiles(templateFiles...)
```

And just have `WithTemplateFiles` panic if it encounters an error.  Or have a final build() call that spits out and previous error? 

Might be slightly better?  I always find methods like `CopyTemplatesToTargetDir ` confusing because the target dir was previously given to the RenderingContext in its "constructor".

paulcwarren (commented):
I put a comment on the code and now it is asking me to submit a review comment...so, just doing that.

jamespollard8 (approved):
LGTM!

tests/ytt/external_blobstore_test.go
davewalter: Yeah I know what you mean. I don't like the idea of intentional panics in code as that would break the tests as opposed to causing them to fail. The `build()` call to check for errors sounds possible, but it seems fragile in that tests could forget to call it.

paulcwarren: Builder pattern is too Java-centric too.  

I think what we need is the [functional options](https://www.calhoun.io/using-functional-options-instead-of-method-chaining-in-go/) pattern.  I have seen this a few times in our code (Credhub is the one I remember).  Perhaps something like this would work better?

```
type RenderingContextOption func(ctx Context) (Context, error)

func NewRenderingContextWithOptions(string, RenderingContextOption...) (Context, error)

ctx, err := NewRenderingContextWithOptions(targetDir,
  WithData(data),
  WithTemplates(templateFiles),
)
Expect(err).NotTo(HaveOccurred())
```

PR #549, opened 2020-10-23
DOCS: add statement to clarify internal database and blobstore usage

paulcwarren:
Clarify that the internal database and blobstore and intended for the kick-the-tyres user only. 

All other users should configure their deployment with external database and blobstore.

[#175406567](https://www.pivotaltracker.com/story/show/175406567)


jamespollard8 (approved):
LGTM!

PR #552, opened 2020-10-27
add readme for package registry

sannidhi:
[#175359589]
[Operator/Platform team should understand how to perform Garbage Collection from the Harbor UI if needed](https://www.pivotaltracker.com/story/show/175359589)

Authored-by: Sannidhi Jalukar <sjalukar@pivotal.io>


acosta11:
Synced up on zoom and confirmed:
- GCR already supports this feasibly for cf
- Docker is feasible and will be supported or documented soon
- Harbor is the known case where support does not yet exist and so cf would like to provide workaround documentation eagerly to operators that may need it

Suggested change before merge:
- Move the contents of the file into the existing system-registry-management doc

Potential cf-for-k8s team follow up change:
- Rename the file itself to `registry-management.md` since it makes more sense to group general registry outcomes in the same doc

Thanks,
Andrew and @ericpromislow

sannidhi:
@acosta11 I moved the package registry deletion docs to the system registry management doc as discussed. Should I also rename the file?

acosta11:
Looks good to me. I think we can worry about the file rename as a team when we do another pass on docs.

Issue #101, opened 2020-03-30
cf cli buildpacks commands are not supported.


Syerram:
## Expected behavior
- When I run `cf buildpacks`, I should get a list of supported buildpacks. 
- I cannot create, update or delete custom buildpacks using `cf [create|update|delete]-buildpacks` commands

## Actual behavior
- When I run `cf buildpacks`, I get an empty result.

## Workaround
None

## Context
This is a known issue and will be tackled post 1.X release. The issue will be updated once we have a timeline.

cwlbraa:
It's worth mentioning that you can also run into this if you do buildpack selection in a cf manifest: https://vmware.slack.com/archives/CSLCWFAG0/p1593632698003900?thread_ts=1593465466.003700&cid=CSLCWFAG0

Syerram:
`cf buildpacks` returns current buildpacks

PR #554, opened 2020-10-28
Fix buildpacks endpoint not accepting zip uploads

cholick:
## WHAT is this change about?
One option for `cf create-buildpack` is to specify a local file. The current nginx CAPI config for this workflow isn't allowing large uploads which results in a `413` when this method is used.

No

## Acceptance Steps
```
cf create-buildpack more-foo9 python-runtime-0.0.169.zip
```

Where `python-runtime-0.0.169.zip` is a zipped up version of a buildpack, for example, this [python buildpack tgz](https://github.com/paketo-community/python-runtime/releases/download/v0.0.169/python-runtime-0.0.169.tgz) uncompressed/untarred and then zipped.

Co-authored-by: Joe Eltgroth <jeltgroth@vmware.com>

### DIFF
--- a/config/capi/_ytt_lib/capi-k8s-release/templates/nginx-configmap.yml
+++ b/config/capi/_ytt_lib/capi-k8s-release/templates/nginx-configmap.yml
@@ -67,6 +67,13 @@ data:
             upload_pass_args on;
             upload_pass_form_field "^resources$";
           }
+
+          location ~ /v3/buildpacks/.*/upload {
+            include public_upload.conf;
+
+            upload_pass_args on;
+            upload_pass_form_field "^resources$";
+          }
       }
 
       server {


cholick:
It looks like maybe uploading buildpacks isn't actually supported yet:

https://github.com/cloudfoundry/cf-for-k8s/issues/101

But this fix will still be needed when it is.

(Getting past the upload, but the next step fails)

acosta11:
Hi @cholick ,

Thanks for the contribution!

> It looks like maybe uploading buildpacks isn't actually supported yet:
> #101
> But this fix will still be needed when it is.

That is correct, cf-for-k8s does not currently support administrative cf buildpack functionality. Given the existing divergence in the kpack model and potential for further divergence of responsibility in capi, I don't know that we want to commit to any cf buildpack related functionality yet.

Additionally, cf-for-k8s vendors in [capi-k8s-release](https://github.com/cloudfoundry/capi-k8s-release), so this change will have to come through that repository.

Thanks,
Andrew and @ericpromislow

PR #556, opened 2020-10-30
DOC: add playbook and info for maintaining the long-running environment

jamespollard8:
This is just a docs change - and I'm hoping to get feedback on it from a couple of teammates.

### DIFF
--- a/docs/maintaining/long-running-environment.md
+++ b/docs/maintaining/long-running-environment.md
@@ -1,0 +1,29 @@
+# cf-for-k8s long-running environment (LRE)
+
+The general idea for our long-running environment is to run it similarly to a production cf-for-k8s, and to measure useful uptime measurements (SLIs) that we can share with stakeholders about the stability of the platform.
+
+## Normal operation
+
+### Nightly Upgrades
+
+The [long-lived-env section of the cf-for-k8s stability pipeline](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s-stability-tests?group=long-lived-env) houses our pipeline jobs that are responsible for deploying and upgrading the LRE. Every weeknight, the environment is automatically updated to the latest HEAD of main.
+
+### SLIs / Recurring Measurements
+
+see https://github.com/cloudfoundry/cf-for-k8s/blob/develop/docs/platform-availability.md#availability-during-normal-operation, which explains the SLIs and includes some SLI results, at least from the weeks leading up to the 1.0 release.
+
+For configuring the Pingdom Uptime check, Release Integration team members have access to https://my.pingdom.com/app/reports/uptime#check=6350471 through the cf-mega@pivotal.io Pingdom account in Lastpass.
+
+## Infrastructure
+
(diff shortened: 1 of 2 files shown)


docs/maintaining/long-running-environment.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Normal operation
+
+### Nightly Upgrades
+
+The [long-lived-env section of the cf-for-k8s stability pipeline](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s-stability-tests?group=long-lived-env) houses our pipeline jobs that are responsible for deploying and upgrading the LRE. Every weeknight, the environment is automatically updated to the latest HEAD of main.
+
+### SLIs / Recurring Measurements
+
+see https://github.com/cloudfoundry/cf-for-k8s/blob/develop/docs/platform-availability.md#availability-during-normal-operation, which explains the SLIs and includes some SLI results, at least from the weeks leading up to the 1.0 release.
```
davewalter, line 13: It feels like this could be a relative link.

docs/maintaining/long-running-environment.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Troubleshooting
+
+In the spirit of the purpose of this environment, we strive to fix the environment using the most tightly-scoped fixes possible. Here are some troubleshooting options in increasingly drastic order:
+
+1) Fix the root problem in cf-for-k8s
+1) Re-push the long-running app using the [deploy-long-lived-node-app job](https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-for-k8s-stability-tests/jobs/deploy-long-lived-node-app)
+1) Re-run the platform upgrade / `kapp deploy` job
+1) Restarting K8s pods
+1) Tearing down and standing back up the environment
```
davewalter, line 29: Can we elaborate on the steps involved here in terms of which CI jobs to run?

davewalter (changes requested):
Hi @jamespollard8,

Great start! I left a couple of small comments, but overall this looks good.

Dave

docs/maintaining/long-running-environment.md
jamespollard8: I was having trouble coming up with any time that I'd want the relative link instead of the very latest. Do you have any examples?

docs/maintaining/long-running-environment.md
jamespollard8: I think it should be as simple as running the destroy job and then the create job, but please fill me in if I'm missing something.

I added a link to the relevant pipeline for convenience

davewalter: I was thinking more about the confusion as to which jobs to run after that to get everything deployed again.

docs/maintaining/long-running-environment.md
davewalter: It feels like good practice to keep all links within the docs relative. If someone happens to be deliberately looking at a particular version of the docs (for example, `main` or a version tag), it seems strange to redirect them to develop in a non-obvious way.

I am also not sure how this will work when we implement a tool to convert the docs for the website.

PR #557, opened 2020-10-31
(minor) fix codefence in deploy instructions

jtigger:
- looks like using roman numerals breaks indented codefences (?)
- using Arabic numerals _does_ work (and delightfully, renders as roman numerals at the 2nd-level)


davewalter:
Thanks @jtigger!

PR #555, opened 2020-10-28
add cf_api_backup_metadata_generator UAA client

tcdowney:
## WHAT is this change about?
This adds a UAA client will be used by the cf-api-backup-metadata-container to scrape some metadata about the CF installation to include as part of a Velero backup.

See this doc for more information:
https://docs.google.com/document/d/1aR6_v0wTrSWpH9G2XqHcUTsdCNzP82ZsiWUxiN-4Zno/edit#heading=h.9fh5dt50qkx0

CAKE Story: [#175469923](https://www.pivotaltracker.com/story/show/175469923)

## Does this PR introduce a change to `config/values.yml`?
Yes

## Acceptance Steps
You can test that this client was successfully created by:

1. Authenticating with UAA / CF API
```
cf auth cf_api_backup_metadata_generator <generated-client-secret> --client-credentials
```

2. Curl some endpoints and see them succeed
```
cf curl /v3/spaces
```

cc/ @cloudfoundry/cf-api-kubernetes-evolution @neil-hickey

PR #559, opened 2020-11-04
fix secret_name for cf_api_backup_metadata_generator client

tcdowney:
## WHAT is this change about?
This is a followup PR for a mistake made in https://github.com/cloudfoundry/cf-for-k8s/pull/555.

It includes a pending change to the cf-api-controllers deployment so that this is adequately tested by CI
It is OK if that change is reverted since it will come through via a vendir sync in the future and works fine without it for now.

That can either be done through CI naturally as it bumps to the latest capi-k8s-release or manually while merging this.

No

## Acceptance Steps
Same as https://github.com/cloudfoundry/cf-for-k8s/pull/555. You should be able to do something like this and see it work after the deploy has succeeded:

```
kubectl exec -n cf-system cf-api-controllers-6cb4d79d5c-z6ntn -c backup-metadata-generator -- "/cnb/process/generate-metadata"
```

CAKE Story: [#175469923](https://www.pivotaltracker.com/story/show/175469923)

### DIFF
--- a/config/capi/capi.yml
+++ b/config/capi/capi.yml
@@ -55,7 +55,7 @@ uaa:
     cf_api_controllers:
       secret_name: cf-api-controllers-client-secret
     cf_api_backup_metadata_generator:
-      secret_name: cf-api-backup-metadata-generator-secret
+      secret_name: cf-api-backup-metadata-generator-client-secret
 
 kpack:
   registry:


davewalter (changes requested):
Hi @tcdowney,

I managed to get the tests running. Looks like you spotted another bug, but that needs to come in via a bump to capi-k8s-release.

Regards,
Dave

tcdowney:
@davewalter I actually was waiting to make the capi-k8s-release change until after this was merged since the deploy doesn't succeed without this cf-for-k8s change. I only included that on the PR cause I hoped it would run through the integration tests and see everything pass. The plan was to remove the `config/capi/_ytt_lib/capi-k8s-release/templates/controllers_deployment.yml` change after that happened.

tcdowney:
I just went ahead and removed it. I tested the deploy locally and it was fine.

config/capi/capi.yml
```diff
@@ -55,7 +55,7 @@ uaa:
     cf_api_controllers:
       secret_name: cf-api-controllers-client-secret
     cf_api_backup_metadata_generator:
-      secret_name: cf-api-backup-metadata-generator-secret
+      secret_name: cf-api-backup-metadata-generator-client-secret
```
tcdowney, line 58: This needs to match this:
https://github.com/cloudfoundry/cf-for-k8s/pull/559/files#diff-8d0ec3e11b92f104c5b1e83ad6bb3d0bfe0d005b976a1c1b31e29af98690093fR127

davewalter:
Sorry about that @tcdowney. Our CI is very protective of the component templates we include in cf-for-k8s. We'll get this merged in once the rest of the tests pass.

tcdowney:
Seems like there is an issue with the external-blobstore tests:
```
Error attaching, falling back to logs: unable to upgrade connection: container minio-client-25f48941698aec50d057e28b not found in pod minio-client-25f48941698aec50d057e28b_default
```

I don't see any reason this would be related, though. 🤔

tcdowney:
@davewalter do those tests need to provide the `--command` flag to `kubectl run`?

```
kubectl run NAME --image=image [--env="key=value"] [--port=port] [--dry-run=server|client] [--overrides=inline-json]
[--command] -- [COMMAND] [args...] [options]
```

https://github.com/cloudfoundry/cf-for-k8s/blob/c07f67382676420ed6b4204fe166c3cc01944e4b/ci/tasks/run-external-blobstore-validation-test/task.sh#L64

davewalter:
@tcdowney We had a misconfiguration in our contributions pipeline for this job. It's green now. :)

Issue #561, opened 2020-11-05
Enable proxy protocol for Ingress Envoy


b1tamara:
**Is your feature request related to a problem? Please describe.**
Some Load Balancers like AWS ELB replace the client IP by their own IP. In that case the feature "X-Forwarded-For" cannot be used because the X-Forwarded-For-Header contains the IP of the Load Balancer. By using Proxy Protocol between LB and Ingress Envoy the client IP will be set in the X-Forwarded-For-Header by Envoy.

**Describe the solution you'd like**
Apply EnvoyFilter for enabling the proxy protocol on the Envoy side. Make it configurable via values.  

```
#@ load("@ytt:data", "data")
#@ if data.values.enable_proxy_protocol:
---
apiVersion: networking.istio.io/v1alpha3
kind: EnvoyFilter
metadata:
  name: proxy-protocol
  namespace: istio-system
spec:
  workloadSelector:
    labels:
      istio: ingressgateway
  configPatches:
  - applyTo: LISTENER
    patch:
      operation: MERGE
      value:
        listener_filters:
        - name: envoy.listener.proxy_protocol
        - name: envoy.listener.tls_inspector
#@ end
```
**Additional context**
The enabling of the proxy protocol of the Load Balancer side is not part of this feature request because it is a Load Balancer specific configuration.

mike1808:
Hi. Can you say why `X-Forwarded-For` header is not a solution? AFAIK Envoy will just add another IP address to that header so it will contain 1. ip of the user 2. ip of the AWS ELB

mike1808:
Also, have you tried configuring `gatewayTopology` config in Istio? https://istio.io/latest/docs/ops/configuration/traffic-management/network-topologies/#configuring-x-forwarded-for-headers

b1tamara:
Hello mike1808,
thank you for your suggested solution. Unfortunately it will not work for LoadBalancers which are not transparent (Transparent indicates whether client IP addresses are visible to the backend servers). 
I enabled the proxy protocol on AWS LB that the Load Balancer preserves the client IP. Then I applied the IstioOperator using the annotation for Gateway istio-ingressgateway:
```
"proxy.istio.io/config": "{\"gatewayTopology\":{\"numTrustedProxies\":1}}",
```
After that I tried to curl the app which returns the "x-Forwarded-For-Header". I got:
```
curl https://x-forwarded-for.<domain>

[...]

PR #564, opened 2020-11-09
Enable proxy protocol for Ingress Envoy

b1tamara:

## WHAT is this change about?
Some Load Balancers like AWS ELB replace the client IP by their own IP. In that case the feature "X-Forwarded-For" cannot be used because the X-Forwarded-For-Header contains the IP of the Load Balancer. By using the proxy protocol between LB and Ingress Envoy the client IP will be set in the X-Forwarded-For-Header by Envoy. This feature must be enabled on both sides. But per default for Envoy the proxy protocol is not enabled. With this change the proxy protocol can be enabled for Envoy by applying an EnvoyFilter.

Yes

## Acceptance Steps
1. Enable the proxy protocol on your LB (not transparent like AWS ELB).
2. Set enable_proxy_protocol value to true.
3. Check that the X-Forwarded-For-Header in a cf app contains the client IP.

@stefanlay

## Things to remember
Feature request: https://github.com/cloudfoundry/cf-for-k8s/issues/561

### DIFF
--- a/config/istio/use-proxy-protocol.yml
+++ b/config/istio/use-proxy-protocol.yml
@@ -1,0 +1,21 @@
+#@ load("@ytt:data", "data")
+#@ if data.values.enable_proxy_protocol:
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
+metadata:
+  name: proxy-protocol
+  namespace: istio-system
+spec:
+  workloadSelector:
+    labels:
+      istio: ingressgateway
+  configPatches:
+  - applyTo: LISTENER
+    patch:
+      operation: MERGE
+      value:
+        listener_filters:
+        - name: envoy.listener.proxy_protocol
+        - name: envoy.listener.tls_inspector
+#@ end
--- a/config/values/00-values.yml
+++ b/config/values/00-values.yml
@@ -77,6 +77,7 @@ use_external_dns_for_wildcard: false
 enable_automount_service_account_token: false
 metrics_server_prefer_internal_kubelet_address: false
 use_first_party_jwt_tokens: false
+enable_proxy_protocol: false
 
 #! configuration for the CF blobstore
 blobstore:
--- a/sample-cf-install-values.yml
+++ b/sample-cf-install-values.yml
@@ -112,6 +112,7 @@ use_external_dns_for_wildcard: true
 enable_automount_service_account_token: true
 metrics_server_prefer_internal_kubelet_address: true
 use_first_party_jwt_tokens: true
+enable_proxy_protocol: true
 
 load_balancer:
   enable: false


stefanlay:
Hi, we just wanted to add the info that we had successfully validated the feature on AWS using an AWS ELB with proxy protocol and an app which logs the X-Forwarded-For header. With the feature the app logged the client IP, without the feature it logged the ELB IP. The same app logs the client IP in a cluster using a transparent LB. We validated this on GCP.

jamespollard8:
Hey @b1tamara, we're struggling to complete the AC steps and we're not sure where we went wrong.

Could you please
1) help be more specific about how to configure the AWS ELB. step-by-step instructions for this would be amazing
2) let us know which app you used that ended up properly providing those headers for you? We used httpbin

Thanks!
James and @acosta11 from the cf-for-k8s maintainers

b1tamara:
Hello,

we added an annotation to the **istio-ingressgateway** service (namespace **istio-system**):
```
kind: Service
metadata:
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-proxy-protocol: '*'
```
 To check the "X-Forwarded-For" Header, send some requests to the cf api or any app and take a look at the log:
```
kubectl logs <istio-ingressgateway-pod> -c istio-proxy -n istio-system
```
In my case I can see it in the logs as the last entry (it does not matter if the route can be found or not):
```
{
"referer":"http://operator.operationsconsole.cf.tboehm-pp.<...>/",
"bytes_sent":"0",
"response_duration":"-",
"upstream_cluster":"-",
"x_b3_traceid":"-",
"x_forwarded_proto":"https",
"downstream_remote_address":"<IP>:50532",
"path":"/","authority":"operator.operationsconsole.cf.tboehm-pp.<...>",
"protocol":"HTTP/1.1","upstream_service_time":"-",
"upstream_local_address":"-","duration":"0",
"upstream_transport_failure_reason":"",
"downstream_local_address":"10.250.34.209:443",
"response_code":"404",
"response_flags":"NR",
"response_tx_duration":"-",
"requested_server_name":"operator.operationsconsole.cf.tboehm-pp.<...>",
"organization_id":"-",
"bytes_received":"0",
"app_id":"-",
"x_b3_spanid":"-",
"process_type":"-",
"x_b3_parentspanid":"-",
"space_id":"-",
"user_agent":"Mozilla/5.0 zgrab/0.x",
"start_time":"2020-11-26T13:56:41.894Z",
"method":"GET",
"request_id":"b656b92a-93a9-4efd-ab91-b12914ad59b4",
"upstream_host":"-",
"x_forwarded_for":"<myClientIP>"
}
```

stefanlay:
When using httpbin, please also add "?show_env=1" to the request, see https://github.com/postmanlabs/httpbin/issues/300

Birdrock:
@b1tamara Hi there! We tried validating on GCP, but ran into some issues. When `enable_proxy_protocol` is set to false, my client IP is returned with `x_forwarded_for`. When we set `enable_proxy_protocol` to true, the connect says it is reset by peer.

What specific configuration steps do I need to take in the GCP console to enable "proxy protocol"? I'd like to validate this in GKE since it is part of or standard development experience, but can also validate in AWS.

axel7born:
I guess "proxy-protocol" is only available for AWS classic load balancers.
https://kubernetes.io/docs/concepts/services-networking/service/#proxy-protocol-support-on-aws
As the gcp loadbalancer doesn't change the source address it's not needed.

b1tamara:
Hi @Birdrock,
The GCP Load Balancer is transparent (Transparent indicates whether client IP addresses are visible to the backend servers). The AWS Classic Load Balancer is not transparent and needs to enable the proxy protocol that the Load Balancer preserves the client IP.

plowin:
Dear community, business cases switched for us and we do not plan to use non-transparent load balancers within 2021.
For us, it would be fine to close this PR and the respective backlog on your end.

If we should get a new requirement which depends on the proxy-protocol, we would raise this again and provide more information on how to reproduce this. What do you think?

jamespollard8:
Sounds good - thanks @plowin for the update. We totally understand.

PR #565, opened 2020-11-10
doc: add transparency to readme banner

XanderStrike:
[Look, look with your special eyes](https://github.com/cloudfoundry/cf-for-k8s/blob/readmeBannerTransparency/README.md)

I think the transparent background looks better on GitHub and definitely will look better in other markdown readers where the background color won't necessarily be known.


jamespollard8:
Great - thanks!

PR #571, opened 2020-11-13
Remove resource requirements also from istio init container

wonderix:
## WHAT is this change about?

If the config variable `remove_resource_requirements` is set to true, the resource requirements should also be removed from the istio init containers and also from `StatefulSets` and `Jobs`.

See also [post in Slack](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1605189845132000)

No

## Acceptance Steps

* Install cf-for-k8s with `remove_resource_requirements=true`
* The resource requirements of all installed pods should be missing or should be zero

@loewenstein, @Haegi , @phil9909


wonderix:
Sorry: opened PR with wrong user.

kramerul:
See #572

PR #572, opened 2020-11-13
Remove resource requirements also from istio init container

kramerul:
## WHAT is this change about?

If the config variable `remove_resource_requirements` is set to true, the resource requirements should also be removed from the istio init containers and also from `StatefulSets` and `Jobs`.

See also [post in Slack](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1605189845132000)

No

## Acceptance Steps

* Install cf-for-k8s with `remove_resource_requirements=true`
* The resource requirements of all installed pods should be missing or should be zero

@loewenstein, @Haegi , @phil9909

### DIFF
--- a/config/istio/remove-resource-requirements.yml
+++ b/config/istio/remove-resource-requirements.yml
@@ -1,13 +1,15 @@
 #@ load("@ytt:data", "data")
 #@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:json", "json")
 
 #@ if data.values.remove_resource_requirements:
 #! remove resource requirements to allow installation on smaller environments
-#@ for/end kind in ["Deployment", "DaemonSet"]:
+#@ for/end kind in ["Deployment", "DaemonSet", "StatefulSet", "Job"]:
 #@overlay/match by=overlay.subset({"kind": kind}),expects="1+"
 ---
 spec:
   template:
+    #@overlay/match missing_ok=True
     metadata:
       #@overlay/match missing_ok=True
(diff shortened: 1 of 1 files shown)


kramerul:
I had a look at the test but was not able to see that the failure is related to the change. The flag `remove_resource_requirements` is not set to `true` for these tests.

jamespollard8 (approved):
Awesome - thanks for this @kramerul!

Comparing the rendered cf-for-k8s yaml before and after the change shows these differences (using `dyff`)

```
spec.template.metadata.annotations (ccdb-migrate job)  (document #171)
  + two map entries added:
    sidecar.istio.io/proxyCPU: 0m
    sidecar.istio.io/proxyMemory: 0M

spec.template.metadata (istio job)  (document #278)
  + one map entry added:
    annotations:
    │ sidecar.istio.io/proxyCPU: 0m
    │ sidecar.istio.io/proxyMemory: 0M

spec.template.spec.containers.cf-db-postgresql  (document #278)
  - one map entry removed:
    resources:
    │ requests:
    │ │ cpu: 250m
    │ │ memory: 256Mi
```

One other [minor] unfortunate thing we saw is that the data.values now show up as an "ugly"/raw json output instead of the pretty one with one line per key but we don't see an easy fix/flag for json.encode


Overall, looks great to us. We're going to fast follow with moving `config/istio/remove-resource-requirements.yml` into the top level config dir - as discussed on slack

cc my pair @acosta11

PR #590, opened 2020-12-03
Re add logging to smoke test

MasslessParticle:
- Make logging happen more frequent to mitigate lossiness

[#175542566](https://www.pivotaltracker.com/story/show/175542566)

Signed-off by: Tim Downey tdowney@vmware.com

## WHAT is this change about?
The smoke test for logging was unrealistic because it only emitted a single log. We made the smoke test app emit 1 log/second to overcome potential lossiness. 

No

## Acceptance Steps
The smoke tests should be less flaky

@tcdowney @angelachin


jamespollard8:
LGTM!

PR #594, opened 2020-12-08
Bump supported k8s versions to include v1.20.0

jamespollard8:
Simply bumping this value should trigger testing it in CI via the `newest-kind` job.


jamespollard8:
Confirmed
`✓ Ensuring node image (kindest/node:v1.20.0)`

PR #599, opened 2020-12-18
ENH: use an annotated minio image

jamespollard8:
- the image came from minio/minio:RELEASE.2020-06-14T18-32-17Z and has
been annotated and moved into the not-rate-limited cloudfoundry
dockerhub

Nope

## Acceptance Steps
Smoke tests (PR tests) pass
and
The minio image used in cf-for-k8s rendered templates comes from cloudfoundry docker hub and has deplab annotations on it when you `docker inspect` the image


Birdrock:
:+1: Looks good.

PR #611, opened 2021-01-14
Add network policy for eirini webhook registration job

kieron-dev:
## WHAT is this change about?

Eirini has split the instance-index-env-injector component into a webhook registration job and a webhook service deployment to facilitate scaling and to remove cluster permissions from the webhook service.

The webhook service requires network to and from the control plane as before, so the eirini-instance-index-env-injector 
network policy is being left as is. The new registration job requires outbound network access to the control plane so we are adding this new network policy targeting the job.

This change is required to support versions of eirini > v2.0.0.

Story: https://www.pivotaltracker.com/story/show/175878489

No

## Acceptance Steps
- Deploy cf-for-k8s with the latest eirini-release from master. See that the cf-system/instance-index-env-injector job completes successfully (if not kapp deployment will fail).
- `cf push` an app. It should succeed and CF_INSTANCE_INDEX should be set in the app's environment. This can be viewed by describing one of the app's pods in cf-workloads.

@cloudfoundry/eirini

### DIFF
--- a/config/networking/network-policies.yaml
+++ b/config/networking/network-policies.yaml
@@ -612,6 +612,21 @@ spec:
   - Ingress
   - Egress
 ---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: eirini-instance-index-env-injector-registrar
+  namespace: #@ system_namespace()
+spec:
+  podSelector:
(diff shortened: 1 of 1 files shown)


jamespollard8 (approved):
Sweet - looks good to us!

PR #613, opened 2021-01-21
Update upgrade-sidecars image digest to sha256:c1eee8fa

jamespollard8:
redo-ing #598 

## WHAT is this change about?
Update the image digest of Istio sidecars upgrade job

## Does this PR introduce a change to config/values.yml?
No

## Acceptance Steps
Deploy succeeds and you can verify that the Istio injected sidecar has the latest version


jamespollard8:
We're actually going to hold off on this until we're updating Istio

PR #615, opened 2021-01-25
Migrate capi-k8s-release pipeline

jspawar:
## WHAT is this change about?
Migrate the [capi-k8s-release pipeline](https://ci.cake.capi.land/teams/main/pipelines/capi?group=capi-k8s-release) into our CI and refactor it a bit to use our pools instead of a single, semi-persistent cluster.

Any unique tasks that pipeline used have been moved into a new directory: `ci/tasks/cake` to make clear they are separate from our more general tasks for cf-for-k8s.

Once this is merged, will have to update the `cf-for-k8s-ci` resource in this pipeline to point at `develop` instead of the branch this PR is coming from.

No

## Acceptance Steps
1. Run `./ci/configure cf-api` and observe that Fly believes there are no changes to apply (since these change have already been applied to the pipeline)


acosta11 (approved):
The port looks reasonable. My only question is whether we want to keep using the capi bot for git commits in the future? Doesn't feel like a blocker on the PR but might want to spin up a tracker story for general gitbot management.

acosta11:
Quick update: it looks like the file extension update wasn't as compatible as I thought. See https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-api/jobs/cc-unit-tests/builds/33 
```
Backend error: Exit status: 500, message: {"Type":"ExecutableNotFoundError","Message":"exec failed: container_linux.go:367: starting container process caused: exec: \"cf-for-k8s-ci/ci/tasks/cake/run_cc_unit_tests\": stat cf-for-k8s-ci/ci/tasks/cake/run_cc_unit_tests: no such file or directory","Handle":"","ProcessID":"","Binary":""}
```

I think we just need to update the task config files to also refer to the scripts with the `.sh` file extension.

Birdrock (approved):
Looks good. Fixed up a few odds and ends with additional commits (branch/file relocation, etc.)

PR #616, opened 2021-01-26
Modify eirini build to use new helmless eirini yaml

mnitchev:
# **Do not apply before eirini release v3.0.0**
This is a draft PR to make sure our changes in the upcoming v3.0.0 release for eirini is compatible with cf-for-k8s. Please do not merge this until that version is released, but please take the time to review it. Our goal for the pure yaml release is to be as convenient as possible for cf-for-k8s to use - this means no overlays will be necessary when building or deploying. To see what these yaml releases look like, follow the [build instructions](https://github.com/cloudfoundry-incubator/eirini-release/tree/11cbbec8554a5f755159a7a9052d9a4ee6e63e34#building-the-yaml-release) on how to render them.

## **Update 12/02/2021**
v3.0.0 has [now been released](https://github.com/cloudfoundry-incubator/eirini-release/releases/tag/v3.0.0) and this PR can now be merged

## WHAT is this change about?

Change build script for eirini  and the templates in config to use the new helmless eirini release.

No

## Acceptance Steps

1. Vendor eirini v3.0.0
1. Run the build script for eirini
1. kapp deploy cf-for-k8s
1. see that everything works as normal

@cloudfoundry/eirini 

## Relevant context:

https://www.pivotaltracker.com/story/show/173896604


jamespollard8 (commented):
Sweet, looking forward to seeing the final thing once the new release is out.

At this point, we likely won't even need the build/eirini script and we can change to directly vendir-ing in Eirini like we do with other components such as capi-k8s-release, uaa, etc

acosta11:
Hi Eirini friends,

Thanks for making the updates to be compatible with the vendir sync checks in CI per our slack discussion in https://cloudfoundry.slack.com/archives/C8RU3BZ26/p1613073589027800. This time we see an error in source app push on KinD and a crash event error in CATs.

Regarding the crash events, the previous configuration in https://github.com/cloudfoundry/cf-for-k8s/commit/0dd4940c144ff1d6ff9ce4a37c1e73050aae1c7e looks slightly different than what we see in the current diff. It looks like the `eirini-events` selector was renamed to `eirini-event-reporter`. So maybe the name/selector mismatch in our separately managed network policies is the issue here.

```
$ ag eirini-events
config/networking/istio-authorization-policies.yml
33:            - #@ principal(system_namespace(), "eirini-events")
config/networking/network-policies.yaml
556:  name: eirini-events
```

We're otherwise not sure about the KinD issue, so hopefully the networking reconciliation will also take care of that.

Thanks,
Andrew and @jamespollard8

danail-branekov:
Thank you for pointing that out, indeed we have overlooked the auth policy yml. This is now fixed.

kieron-dev:
> This time we see an error in source app push

Regarding this, we also see this in kind. The pod in cf-workloads-staging shows the problem described in issue https://github.com/cloudfoundry/cf-for-k8s/issues/444.

We upgraded kind to the latest 0.10.0, but the problem persisted. When we applied the suggested patch from the issue to `deploy/kind/cluster.yml`, this fixed the problem. So it seems that the cb835ec7490f5ab068083d04d440ee28898c1cb9 commit reverting that patch is a little premature.

jspawar:
@kieron-dev yes it looks like the commit to remove that patch was a bit too optimistic and we've reverted that.

We however cannot re-trigger the PR checks because it looks like there is a merge conflict we were hoping you could resolve first please. 

From our end, I believe we made changes to this `rendered.yml` file to address a separate issue you all caught about the version mismatch of your images we're using: https://github.com/cloudfoundry/cf-for-k8s/pull/633

gcapizzi:
Hey team! We _think_ this is ready but:
1. How do you test your PRs? We've some brief manual testing but couldn't run CATs, how do we get enough confidence that this works?
2. `osl-compliant-image-override.yml` still replaces our images with your builds which I guess are still pointing at 2.0 - I guess you are going to build new OSL-compliant images for 3.0 and update that file yourselves?

Thanks!
/cc @cloudfoundry/eirini

gcapizzi:
Yeah it looks like smoke tests are failing ☹️ We'll take a look, bear with us!

gcapizzi:
Hey @jamespollard8 @acosta11 @jspawar! I just got the smoke tests to pass on my machine, and looking at the output of the failing builds ([here](https://release-integration.ci.cf-app.com/builds/729259) and [here](https://release-integration.ci.cf-app.com/builds/729260)) I can't really tell what the problem is 😞  can you help?

acosta11:
Hi @gcapizzi ,

Thanks for the updates and the quick follow up on the smoke test issue! From [the slack thread](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1615569342085500), I think I see what the issue is on KinD now:

I was able to reproduce locally on KinD w/ k8s v16.15 as we use in CI.

It looks like the app was hitting an image pull backoff issue:
```
Warning  Failed     2m40s (x3 over 3m24s)  kubelet            Failed to pull image "<app image>": rpc error: code = Unknown desc = failed to pull and unpack image "<app iamge>": failed to resolve reference "<app image>": unexpected status code [manifests <sha256 digest>]: 401 Unauthorized
  Warning  Failed     2m40s (x3 over 3m24s)  kubelet            Error: ErrImagePull
  Normal   BackOff    2m14s (x4 over 3m23s)  kubelet            Back-off pulling image "<app image>"
  Warning  Failed     2m14s (x4 over 3m23s)  kubelet            Error: ImagePullBackOff
```

And then I saw that there weren’t any image pull secrets on pod spec, nor was there any reference to an app registry secret in any of the service accounts in use in the workloads namespace. It looks like there is a reference mismatch between the opi configmap and the secret created in cf-for-k8s for it to consume.

`config/eirini/eirini.yml` in cf-for-k8s
```
apiVersion: v1
kind: Secret
metadata:
  name: app-registry-credentials
```

`config/eirini/_ytt_lib/rendered.yml`  in cf-for-k8s from the output of helm template on eirini release
```
apiVersion: v1
data:
  opi.yml: |
    app_namespace: cf-workloads
     ...
    registry_secret_name: registry-credentials
```

After syncing those references up, the node app push succeeded for me.

gcapizzi:
Thank you @acosta11! My local setup is using a public Docker registry, I think that's why I'm not seeing it fail locally 🙁
I've just pushed a fix 🤞  As mentioned [here](https://github.com/cloudfoundry/cf-for-k8s/pull/616#issuecomment-797516812) you'd still have to update the OSL image references but other than that, I think we're good!

gcapizzi:
Ok `external-db-rds` is failing now, although it used to pass, so maybe a flake?

jspawar:
@gcapizzi appears to have been a flake, everything is all green now though so we're going to go ahead and merge! Thanks for all the help and patience getting this through y'all

PR #625, opened 2021-02-17
Generate the instance index env injector certificate

danail-branekov:
## WHAT is this change about?
With the next Eirini release we are dropping EiriniX as a framework for our instance index environment variable injector (the one that injects CF_INSTANCE). Instead we are managing the webhook configuration statically via a [dedicated yaml](https://github.com/cloudfoundry-incubator/eirini-release/blob/b0dc988c4a4ff53b15dcbe040b7ada5be1f41756/helm/templates/core/instance-index-env-injector-webhook.yml) in the eirini deployment yamls.

In order to adopt this change in CF4K8S needs to 
* generate the webhook certificate bundle into a K8S secret which is mounted into the Eirini instance index env injector
* set the `caBundle` property to the base64 encoded webhook CA bundle into the `MutatingWebHookConfiguration`

This is what this PR does.

**Please note** that dropping EiriniX results into the webhook created by EiriniX (`eirini-x-mutating-hook`) to no longer work thus preventing statefulsets from creating pods (i.e. `cf push` would be broken). Therefore its `MutatingWebHookConfiguration` should be deleted as part of the cf-for-k8s upgrade process. Is there an automated step to achieve this?

Yes

## Acceptance Steps
1. Deploy CF4K8S
2. Push an application
3. ssh onto the application pod via `kubectl exec -n cf-workloads --container opi -it <application-pod-id> /bin/bash`
4. Once in the shell, verify that the `CF_INSTANCE_INDEX` env var is set via `echo $CF_INSTANCE_INDEX`

@cloudfoundry/eirini 

## Things to remember
* Eirini story: https://www.pivotaltracker.com/story/show/176823153

### DIFF
--- a/config/eirini/eirini.yml
+++ b/config/eirini/eirini.yml
@@ -10,6 +10,17 @@
 #@ eirini = library.get("eirini")
 --- #@ template.replace(eirini.eval())
 
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: instance-index-env-injector-certs
+  namespace: #@ system_namespace()
+stringData:
+  tls.crt: #@ data.values.instance_index_env_injector_certificate.crt
+  tls.key: #@ data.values.instance_index_env_injector_certificate.key
+  tls.ca: #@ data.values.instance_index_env_injector_certificate.ca
+
 #! Allow app traffic from the istio-ingressgateway
 ---
 apiVersion: networking.k8s.io/v1
--- a/config/eirini/set-instance-index-env-injector-webhook-ca.yml
+++ b/config/eirini/set-instance-index-env-injector-webhook-ca.yml
@@ -1,0 +1,12 @@
+#@ load("@ytt:overlay", "overlay")
+#@ load("@ytt:data", "data")
+#@ load("@ytt:base64", "base64")
+
+#@overlay/match by=overlay.subset({"kind": "MutatingWebhookConfiguration", "metadata":{"name": "eirini-x-mutating-hook"}})
+---
+webhooks:
+#@overlay/match by="name"
+- name: instance-index-env-injector.eirini.cloudfoundry.org
+  clientConfig:
+    #@overlay/replace
+    caBundle: #@ base64.encode("{}".format(data.values.instance_index_env_injector_certificate.ca))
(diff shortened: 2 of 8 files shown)


jamespollard8:
Hey @danail-branekov, thanks for the PR and for including tests changes.

1) Why is the new certificate necessary? Could we just use Istio service mesh to handle mTLS?

2) 
> `MutatingWebHookConfiguration` should be deleted as part of the cf-for-k8s upgrade process. Is there an automated step to achieve this?

Unfortunately we're not aware of any automation options here, and we'd strongly prefer to implement this without a breaking change if possible.

Best,
James and @jspawar

kieron-dev:
> 1. Why is the new certificate necessary? Could we just use Istio service mesh to handle mTLS?

The cert is for the cluster control plane -> webhook service communication. It actually plain TLS with no client authentication, which should be fine for just mutating yaml strings.

I _think_ it's not possible for the control plane to successfully call to a service with istio wrapping. The control uses `<service-name>.<namespace>.svc` as the hostname in the request, and expects that to be included in the certificate. But when we last experimented with this leaving istio involved, we saw istio inserted some strange host names in its certificate, and the expected name was absent. So we turned off istio of the webhook service deployment and manually provided a certificate.

We believe other admission webhooks go for a similar approach, but create the hook and certificate dynamically, which has the downside of leaving permission to register admission webhooks in runtime. So we've gone for the safer approach of creating the hook as part of the deployment step.

danail-branekov:
We have updated the PR to make the webhook reuse the name of the EiriniX version (`eirini-x-mutating-hook`). This is the easiest way to address the Eirini bump issue described in the PR description. Also see [this slack discussion](https://cloudfoundry.slack.com/archives/CH9LF6V1P/p1614099287007700) for details.

matt-royal:
Hi @danail-branekov. It looks like the unit tests are still failing due to config/eirini/set-instance-index-env-injector-webhook-ca.yml. It assumes that the eirini-x-mutating-hook will have a webhooks key, but it doesn't appear to. Please fix the failing test and then we'll merge.

gcapizzi:
Closing in favour of #643.

Issue #614, opened 2021-01-21
"info.Labels: label key and value greater than maximum size" on docker pull


joscha-alisch:
Unfortunately we seem to be hitting a similar bug to this one: #444 but on GKE instead of KinD.

We run cf-for-k8s version 1.1.0 on GKE version `1.18.12-gke.1201` with nodes running on image type `cos_containerd`.

When we push an app via `cf push`, we see the following error:

```
Failed to pull image "gcr.io/PROJECT_ID/cf-workloads/cf-default-builder@sha256:b65a50427c08b4ade4e75582a4d6fa86c6da0eec7c1f1932c1881154c3527fb5": 
[rpc error: code = InvalidArgument desc = failed to pull and unpack image "gcr.io/PROJECT_ID/cf-workloads/cf-default-builder@sha256:b65a50427c08b4ade4e75582a4d6fa86c6da0eec7c1f1932c1881154c3527fb5": failed to prepare extraction snapshot "extract-929517039-zaeh sha256:e693876ebf739f21944936aabae94530f008be3cb7f14f66c4a2f4fd9b4bcf54": 
info.Labels: label key and value greater than maximum size (4096 bytes), key: containerd: invalid argument, rpc error: code = InvalidArgument desc = failed to pull and unpack image "gcr.io/PROJECT_ID/cf-workloads/cf-default-builder@sha256:b65a50427c08b4ade4e75582a4d6fa86c6da0eec7c1f1932c1881154c3527fb5": failed to prepare extraction snapshot "extract-809081699-XqHo sha256:e693876ebf739f21944936aabae94530f008be3cb7f14f66c4a2f4fd9b4bcf54": 
info.Labels: label key and value greater than maximum size (4096 bytes), key: containerd: invalid argument]
``` 

When we remove buildpack-groups from the cf-default-builder in `cf-for-k8s/config/kpack/default-buildpacks.yml` then it works fine (only keeping the go-buildpack for example). So the assumption is, that these cause too much metadata in the resulting image.

jamespollard8:
Oh interesting - thanks @joscha-alisch for filing this issue!

From what I remember from #444, it looked like this issue was going to be fixed by an update in containerd. So hopefully that'll get into one of the next GKE releases / `cos_containerd`.

Until then, maybe you'll need to use the default cos (Docker) for your GKE cluster.

Does that sound reasonable?

jspawar:
[...]

PR #629, opened 2021-02-22
MAINT: remove KinD patch to disable snapshot annos

jamespollard8:
- Kind should now be using a new enough version of containerd (v1.4.2+) such that this work-around isn't required
- Related to #444

## Acceptance Steps
PR tests pass, specifically the KinD tests


jspawar:
GKE-related checks failed, but only because we hadn't introduced our fix related to: #614 

The Kind-related checks passed, so we're reasonably convinced this is good to merge and doing so as a result

PR #643, opened 2021-03-19
Bump Eirini to v3.1.0

gcapizzi:
## WHAT is this change about?
* Bump Eirini to v3.1.0 in `vendir.yml`
* Adopt restructuring of Eirini config maps: In order to support different configurations in the CRD flow, ConfigMaps have been split into two (one for the traditional REST API and one for the CRD flow). This change adopts that change and sets the `unsafe_allow_automount_service_account_token` to `true` in both REST and CRD configurations.
* Generate the instance index env injector certificates: as the instance index env inject is not based on Eirinix anymore, its certificates need to be generated at deploy time.
* Add the necessary NetworkPolicy to make the eirini-app-migration Job work.

Yes, to `config/values/20-secrets-config-values.yml`.

@cloudfoundry/eirini


gcapizzi:
Hey team! Tests were failing due to a missing `NetworkPolicy`, and adding it made smoke tests passing locally, not sure why the same failure is persisting in some of the checks :/

Birdrock:
Thanks @gcapizzi! We'll take a look at this and try to get it out in a release in the near future.

build/eirini/_vendir/eirini/core/migration-job.yml
```diff
@@ -0,0 +1,33 @@
+---
```
jspawar, line 1: What is this job doing?

config/networking/network-policies.yaml
```diff
@@ -721,3 +721,20 @@ spec:
...
+metadata:
+  name: eirini-app-migration
+  namespace: cf-system
+spec:
+  podSelector:
+    matchLabels:
+      name: eirini-app-migration
+  egress:
+  - {}
+  ingress:
```
jspawar, line 736: Why is ingress being allowed?

jspawar (commented):
Seems okay, but we would love if you could clarify what exactly the `eirini-app-migration` job is doing and how? Some questions for instance, is this job intended to run only for this particular upgrade or for future upgrades as well?

build/eirini/_vendir/eirini/core/migration-job.yml
gcapizzi: We have effectively implemented a migration framework for Eirini StatefulSets, and this job triggers it. At the moment we have one migration in place, it's needed to retroactively fix a bug in the way Eirini calculated CPU requests, which lead to very inefficient scheduling of workloads on nodes. We'd like the Job to run at every upgrade, and it's designed to only apply the migrations that are missing for any particular StatefulSet.

config/networking/network-policies.yaml
gcapizzi: I think this is a mistake, I'm pretty sure we don't need it. I'll fix it ASAP.

gcapizzi: Done!

build/eirini/_vendir/eirini/core/migration-job.yml
Birdrock: @jspawar Any further concern here?

tcdowney: Jwal and I reviewed this together yesterday and this answers the questions we had. Thanks @gcapizzi.

My only other ask is that that context about the migration Job be captured in the "Eirini is composed of:" of its [README](https://github.com/cloudfoundry-incubator/eirini/blob/master/README.md), but that shouldn't block this PR.

Birdrock: @tcdowney @jspawar Thanks! We can move this forward.

PR #657, opened 2021-04-15
Bump Eirini to v4.0.0

gcapizzi:
## WHAT is this change about?
This bumps Eirini to `v4.0.0` and applies some changes that are needed or possible with this new release.

In particular:
* `build/eirini/overlays/disable-cc-api-tls.yml` is not needed anymore as that's now the default;
* the `opi` image has been renamed to `api`;
* the `resource-validator` component has been introduced;
* the `registry-credentials` `Secret` has been renamed back to `app-registry-secret` so `config/eirini/rename-registry-credentials.yml` is not needed anymore;
* the `args` overrides in `config/eirini/image-overrides.yml` are no longer necessary.

This passes the smoke tests if I point `config/values/10-eirini-images.yml` to our [latest images](https://github.com/cloudfoundry-incubator/eirini-release/blob/399b57683e39f69e47d3cbb958d64c4823225a97/helm/values.yaml#L93-L100).

`config/values/10-eirini-images.yml` changes the `opi` key to be `api`.

@cloudfoundry/eirini


gcapizzi:
I _think_ the failures above are due to the images not being up-to-date.

heycait:
Hmm.. since it seems to be an issue with the images, could we update those first and merge it in and re-trigger the tests for this PR? Or do you think the changes to the image getting merged first will break what is currently in cf-for-k8s? 

Perhaps this doesn't matter either way because this will be a fast follow up... I was asking to see if we should create another story to address the image bumps first and then let this one go through again.

gcapizzi:
> Hmm.. since it seems to be an issue with the images, could we update those first and merge it in and re-trigger the tests for this PR? Or do you think the changes to the image getting merged first will break what is currently in cf-for-k8s?

Yeah I think the new images will break in current cf-for-k8s. The ideal thing to do would be to update this PR with the new images, but I don't know if you already have them built and which ones they are, is there a way for me to find out?

heycait:
After speaking with the rest of the team, it sounds like we'll want to first merge in this PR https://github.com/cloudfoundry/cf-for-k8s/pull/658 and have you rebase your changes on top of that and update image references where needed. 

This sounds like a temporary issue until we're done with our image building work, so for this PR, you can go ahead and find all references to Eirini images and replace them with the updated ones ya'll have (and then we'll follow up later to rebuild them).

gcapizzi:
Sounds good, we should be green as soon as we rebase on top of #658! We should probably also think of a future in which we just build the right images to begin with, maybe by using your tooling?

davewalter:
@gcapizzi PR #658 has been merged so you should just be able to rebase on HEAD of develop.

gcapizzi:
@davewalter done! it looks like the build is stuck due to some terraform issue though ☹️

jspawar:
Yea that appeared to have been a strange flake, has since been resolved. This looks good to go then, so merging. Thanks!

PR #658, opened 2021-04-15
Remove image override data values

matt-royal:
## WHAT is this change about?
Remove image overrides for eirini, istio, minio, and postgres.
  - Remove config/values/*.yml files that set image overrides to use
  - Remove ytt overlays that overrode the images
  - 
For quarks, revert to older solution that used kbld to do overrides.
This solution is superior, because it won't apply the override if the
quarks image tag changes.
This was done by reverting (-n) 6f2122925b81c85277b3d78432b096176521dc46

## Does this PR introduce a change to `config/values/`?
Yes

## Acceptance Steps
No feature changes... it should still deploy and run

@acosta11 

Here's the story: https://www.pivotaltracker.com/story/show/177786619


Birdrock:
Merging this now - we are considering the direction for all components and this adds complexity. Will revisit when we have fleshed out the internal component mechanisms.

PR #664, opened 2021-05-04
Update the holdApplicationUntilProxyStarts Istio setting

davewalter:
## WHAT is this change about?
When we run the `build/istio/build.sh` script with istioctl 1.9.4, we saw a deprecation warning about the `holdApplicationUntilProxyStarts` configuration:
```
$ ./build/istio/build.sh
generating Istio resource definitions...
! values.global.proxy.holdApplicationUntilProxyStarts is deprecated; use meshConfig.defaultConfig.holdApplicationUntilProxyStarts instead
resolve | final: cloudfoundry/cf-k8s-networking-fluentbit@sha256:2b8563a032c007bdb5f8a1cdbd093edd8505fadd21a3443a2b891e87caee7504 -> index.docker.io/cloudfoundry/cf-k8s-networking-fluentbit@sha256:2b8563a032c007bdb5f8a1cdbd093edd8505fadd21a3443a2b891e87caee7504
resolve | final: index.docker.io/istio/pilot:1.9.4 -> index.docker.io/istio/pilot@sha256:a613e4535cad4028d0aca7e16450f16f41ed2466fa522f6c22c2b96bfbc26eba
resolve | final: index.docker.io/istio/proxyv2:1.9.4 -> index.docker.io/istio/proxyv2@sha256:5e5d6153164c6a71fb1538cb11b3b8c13edc7f9628a4927c180df7ea5c02bae2
```

This PR updates the Istio configuration to use the new property.

[#178029864](https://www.pivotaltracker.com/story/show/178029864)

No

## Acceptance Steps
Deploy as normal.

@Birdrock

### DIFF
--- a/config/istio/istio-generated/xxx-generated-istio.yaml
+++ b/config/istio/istio-generated/xxx-generated-istio.yaml
@@ -4494,6 +4494,7 @@ data:
       }
     defaultConfig:
       discoveryAddress: istiod.istio-system.svc:15012
+      holdApplicationUntilProxyStarts: true
       proxyMetadata: {}
       tracing:
         zipkin:
@@ -5308,7 +5309,7 @@ data:
           "excludeIPRanges": "",
           "excludeInboundPorts": "",
           "excludeOutboundPorts": "",
-          "holdApplicationUntilProxyStarts": true,
+          "holdApplicationUntilProxyStarts": false,
           "image": "proxyv2",
           "includeIPRanges": "*",
           "logLevel": "warning",
--- a/build/istio/istioctl-values.yaml
+++ b/build/istio/istioctl-values.yaml
@@ -76,6 +76,5 @@ spec:
           "x_forwarded_for": "%REQ(X-FORWARDED-FOR)%",
           "x_forwarded_proto": "%REQ(X-FORWARDED-PROTO)%"
         }
-    global:
-      proxy:
+      defaultConfig:
         holdApplicationUntilProxyStarts: true


config/istio/istio-generated/xxx-generated-istio.yaml
```diff
@@ -5308,7 +5309,7 @@ data:
           "excludeIPRanges": "",
           "excludeInboundPorts": "",
           "excludeOutboundPorts": "",
-          "holdApplicationUntilProxyStarts": true,
+          "holdApplicationUntilProxyStarts": false,
```
tcdowney, line 5312: It's weird that the generated config becomes false here. What you configured looks correct, though, so I assume this still works. 👍

tcdowney:
For context, we set the `holdApplicationUntilProxyStarts` property to support apps that need to be able to make network calls as part of start up. If the sidecar proxy isn't ready when these apps try to start they will crash.

I think if we want to check that there are no functional regressions we could push the app referenced in [this comment](https://www.pivotaltracker.com/story/show/174125593/comments/218256224). If the `cf push` succeeds for it then it means that the app container wasn't started until the proxy was ready.

tcdowney (approved):

tcdowney:
I created a chore to add a test to the cf-k8s-networking acceptance tests so that we can get some automated validation of this: https://www.pivotaltracker.com/story/show/178040917

tcdowney:
@acosta11 and I manually validated this by deploying the changes from this PR and running the test we added in the story above (https://github.com/cloudfoundry/cf-k8s-networking/commit/9726f4997c57c8cf250a345ac7e19c89dbd570a3).

Issue #663, opened 2021-04-30
CF does not rollout sidecar updates to internal workloads


braunsonm:
## Describe the bug
See this architecture decision for context: https://github.com/cloudfoundry/cf-k8s-networking/blob/35f3ed39e26293aef16a5afae36d7a5271c13c77/doc/architecture-decisions/0016-job-for-upgrading-istio-sidecars-on-workloads.md

Currently when a new istio version is released the control plane is updated first, then all `cf-workloads` pods are restarted with the new sidecar.

This does not apply to `cf-system` and `cf-blobstore` workloads. Istio only guarantees that sidecars will be forward compatible with 1 minor revision. If a component is not updated in CF updates between then it could begin running a version older than the control plane supports.

This has recently happened to all operators in the move from `v3.0.0` -> `v4.0.0` -> `v4.1.0`. 

So far this hasn't caused any issues but could in the future and is not considered good practice to run outdated sidecars. 

You can also verify in the istio docs their last step of the upgrade process instructs you to restart all data plane workloads: https://istio.io/latest/docs/setup/upgrade/in-place/

## To Reproduce*
1. Deploy v3
2. Upgrade to v4.0.0
3. Upgrade to v4.1.0
4. Notice that cf-blobstore is now running a sidecar that is unsupported by the istio control plane. (v1.7.3)

## Expected behavior
All CF components should be restarted after a control plane update to istio for compatibility. 

The job for this should be tweaked to rollout the update to move than just cf-workloads and fluentd: https://github.com/cloudfoundry/cf-k8s-networking/blob/35f3ed39e26293aef16a5afae36d7a5271c13c77/ci/dockerfiles/upgrade/roll.sh#L15

I would be curious to hear from the networking team for the rationale in only restarting fluentd.

### cf-for-k8s SHA 
v4.1.0

### Cluster information
AKS

acosta11:
Hi @braunsonm ,

Thanks for the issue with the relevant networking repo context. For visibility, we have a story in the backlog to investigate solutions to potentially undesirable upgrade deployment behavior. See https://www.pivotaltracker.com/story/show/177914742 .

[...]

PR #665, opened 2021-05-04
ENH: Add an Istio version label to all system components

davewalter:
## WHAT is this change about?
This PR adds an Istio version label to all system components. This will cause kapp to roll the components whenever the Istio version changes, which will result in the new version of the proxy sidecar being injected into the pods.

[#177914742](https://www.pivotaltracker.com/story/show/177914742)

No

## Acceptance Steps
Deploy a previous version of cf-for-k8s that includes an old version of Istio, then upgrade to this version and confirm that all system component pods are rolled during the upgrade, and that kapp waits for them all to be ready again before proceeding.

@Birdrock


acosta11:
We manually verified that everything rolled on an update from Istio 1.9.4 -> 1.9.5 as expected. LGTM

Thanks,
Andrew and @tcdowney

Birdrock:
fixes #663

PR #669, opened 2021-06-06
Update vendir.yml

BigGuy573:


BigGuy573:
@cf-rel-int-status-bot fix for #669 @ `[](https://release-integration.ci.cf-app.com/builds/835337) `

